PR #50, opened 2024-10-16
Implement React Native framework-less to design-system-react-native package

brianacnguyen:
## **Description**
This PR implements React Native framework-less to design-system-react-native package

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-design-system/issues#workspaces/design-system-61e8a2ae77c3a60012e5003c/issues/zh/809

## **Manual testing steps**

1. Run `yarn install`
2. Go to `/packages/design-system-react-native/ios/` ==> Run `pod install`
3. From root, run `yarn workspace @metamask/design-system-react-native start`
4. Go to `/packages/design-system-react-native/ios/` ==> Open `MetamaskDesignSystemReactNative.xcworkspace` with Xcode ==> Press Play

### **After**
Simulator Screenshot - iPhone 16 Pro - 2024-10-16 at 14 24 59

- [ ] I’ve included tests if applicable


packages/design-system-react-native/package.json
```diff
@@ -39,22 +34,47 @@
...
+    "start": "react-native start",
+    "test": "jest --reporters=jest-silent-reporter",
+    "test:clean": "jest --clearCache",
+    "test:verbose": "jest --verbose",
+    "test:watch": "jest --watch"
+  },
+  "dependencies": {
+    "react": "18.2.0",
+    "react-native": "0.72.0"
   },
```
garrettbear, line 51: should be in devDependencies and peerDependencies (in peer it should not be a strict version e.g. 18.0.0)

package.json
```diff
@@ -44,8 +45,13 @@
   "resolutions": {
     "elliptic@6.5.4": "^6.5.7",
     "fast-xml-parser@^4.3.4": "^4.4.1",
+    "react-native": "^0.75.4",
     "ws@7.4.6": "^7.5.10"
   },
+  "dependencies": {
+    "react": "^18.3.1",
+    "react-native": "^0.75.4"
+  },
```
georgewrmarshall, line 54: Should these be in devDependencies?

georgewrmarshall (commented):
Hey @brianacnguyen, this is looking like a great start! I just have a few questions that may be due to my lack of knowledge of React Native.

1. Metro bundler: Given that we're creating UI components rather than a full app, do we need the full Metro setup? Would this be more appropriate in the docs/storybook-reactnative folder? We might be able to simplify our dev dependencies and scripts.

2. Dependencies: I noticed we're including React and React Native as direct dependencies both in the root monorepo and in design-system-react-native package. Since we're building a component library, we might want to consider moving these to `devDependencies` or `peerDependencies` instead. This would allow more flexibility for projects using our components.

3. node_modules: I saw we're ignoring `node_modules` in the package's ESLint config. With Yarn workspaces, we typically don't need package-level `node_modules`. Is there a specific reason we're including this?

4. React compatibility: We're planning to add a `design-system-react` package soon. How do you think we should manage shared dependencies like React across these packages?

package.json
garrettbear: should be in devDependencies and peerDependencies

PR #89, opened 2024-11-08
chore: updating migration docs to remove dependabot step

georgewrmarshall:
## **Explanation**

This PR removes the section regarding disabling `dependabot` alerts and updates from the migration guide documentation, based on organizational requirements discussed in this [Slack thread](https://consensys.slack.com/archives/C01V1L10W2E/p1731097509734129). It was determined that this step is no longer necessary because:

- Dependabot alerts cannot be disabled at the repository level due to new organizational rules.
- Disabling Dependabot alerts would ultimately be redundant, as the repository will be archived after the migration, which will disable Dependabot by default.
- This step was originally intended to prevent confusion due to updates after migration began, but archiving the repository will achieve the same effect once complete.

## **References**

- Related conversation in Slack: [Discussion on Dependabot configuration](https://consensys.slack.com/archives/C01V1L10W2E/p1731097509734129).

### DIFF
--- a/docs/package-migration-process-guide.md
+++ b/docs/package-migration-process-guide.md
@@ -12,15 +12,9 @@ This document outlines the process for migrating a MetaMask library into the met
 
 - [Example PR](https://github.com/MetaMask/eth-json-rpc-provider/pull/38)
 
-### 2. Disable `dependabot` dependency version updates and security alerts
+### 2. Add the source repo to the ZenHub workspace repo filter so that its issues/PRs show up on the board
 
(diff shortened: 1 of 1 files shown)


docs/package-migration-process-guide.md
```diff
@@ -12,35 +12,29 @@ This document outlines the process for migrating a MetaMask library into the met
...
 
 - Modify the "name" field in `package.json`.
 - Update the title of the README.md.
 - Add a CHANGELOG entry for the rename.
 
-### **[PR#5]** 7. Create a new release of the migration target from the source repo
+### **[PR#5]** 6. Create a new release of the migration target from the source repo
 
-- All subsequent releases of the migration target will be made from the metamask module template.
+- All subsequent releases of the migration target will be made from the metamask design system monorepo.
```
georgewrmarshall, line 37: Fixing mistake reference to module template

brianacnguyen (approved):

Issue #316, opened 2025-01-08
[React Native]: Create`TextButton` component


georgewrmarshall:
### **Description**

Create a new `ButtonText` component in React Native, designed to act as a text-styled navigational button. This component should follow the specifications outlined in the [[ButtonText Audit and Insight Report Figma Board](https://www.figma.com/board/SEQligIgoo7fWYGYJ9uJRJ/Button%2FHyperlink-Audit-and-Insight-Report?node-id=0-1&t=HUJUu6JSQjf4tXpx-1)](https://www.figma.com/board/SEQligIgoo7fWYGYJ9uJRJ/Button%2FHyperlink-Audit-and-Insight-Report?node-id=0-1&t=HUJUu6JSQjf4tXpx-1). It should prioritize accessibility, responsiveness, and consistency with the shared design system.

### **Technical Details**

- **Component Implementation**:
  - Implement a text-styled button component adhering to design tokens and visual guidelines.
  - Include support for standard button attributes (e.g., `onPress`, `accessibilityLabel`, `testID`).
  - Ensure semantic use of React Native components for proper accessibility and keyboard navigation.

- **Styling**:
  - Apply consistent design tokens for typography, spacing, and color.
  - Include hover, focus, and active states as defined in the Figma board.
  - Ensure responsiveness and compatibility with the shared UI library across multiple screen sizes.

- **Integration**:
  - Ensure the component integrates seamlessly with the shared design system.
  - Validate compatibility with design tokens for colors, typography, and spacing.

- **Testing**:
  - Write unit tests to cover key functionalities, including edge cases.
  - Include accessibility tests to verify keyboard navigation, focus management, and screen reader compatibility.

- **Documentation**:
  - Add a Storybook entry for the `ButtonText` component:
    - Showcase examples for default, hover, and focus states.
    - Provide variations for use cases such as primary actions and secondary actions.
  - Include usage guidelines and best practices for integration in the component library documentation.

### **Acceptance Criteria**

[...]

PR #381, opened 2025-01-28
[DSRN] Add TextButton

brianacnguyen:
## **Description**
This PR adds the `TextButton`

## **Related issues**

Fixes: #316 

## **Manual testing steps**

2. Go to Components > TextButton
3.

### **After**
https://github.com/user-attachments/assets/8833a195-134d-4be4-a756-420cd0b5ccc0


packages/design-system-react-native/src/components/TextButton/TextButton.constants.ts
```diff
@@ -0,0 +1,43 @@
...
+    size: IconSize.Sm,
+    testID: 'start-icon',
+  },
+  endIconProps: {
+    size: IconSize.Sm,
+    testID: 'end-icon',
+  },
+  onPress: () => {
+    console.log('Button pressed');
+  },
```
georgewrmarshall, line 40: We shouldn't use a console.log as default props. You could consider using a no-op function (`() => {}`) or removing the default `onPress` handler entirely.

packages/design-system-react-native/src/components/TextButton/TextButton.tsx
```diff
@@ -0,0 +1,243 @@
...
+    });
+  }, [isPressed, isLoading, isDisabled, twClassName]);
+
+  const twTextClassNames = useMemo(() => {
+    return generateTextButtonTextClassNames({
+      isPressed,
+      isInverse,
+      isLoading,
+    });
+  }, [isPressed, isInverse, isLoading, textProps]);
```
georgewrmarshall, line 63: Should `textProps` be removed from the dependency array here? It doesn't seem to be used in the `twTextClassNames` useMemo hook. Not sure why linting didn't pick this up 🤔

georgewrmarshall (commented):
LGTM! Great test coverage and documentation. Unfortunate about the issues with React Native and the need for nested `View` and empty `Text` JSX elements. Left some non-blocking comments.

georgewrmarshall (approved):

packages/design-system-react-native/src/components/Icon/Icon.types.ts
```diff
@@ -55,15 +55,15 @@ export enum IconColor {
...
+  Sm = '16',
   /** Medium - 20px (Default) */
-  Md = 'md',
+  Md = '20',
   /** Large - 24px */
-  Lg = 'lg',
+  Lg = '24',
   /** Extra large - 32px */
-  Xl = 'xl',
+  Xl = '32',
```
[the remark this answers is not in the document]
brianacnguyen: in my experience, using these is a lot easier with Tailwind since I can just use 
`h-[${IconSize}]` in places that I need to use the `IconSize` to style or calculate instead of adding a lot of different mappings. wdyt?

packages/design-system-react-native/src/components/TextButton/TextButton.tsx
brianacnguyen: good catch! it was used at some point for the calculation but is no longer needed so I'll remove it

georgewrmarshall (approved):

PR #415, opened 2025-02-04
[DSRN] Add ButtonPrimary 

brianacnguyen:
## **Description**
This PR adds the `ButtonPrimary` component to the DSRN

## **Related issues**

Fixes: #177 

## **Manual testing steps**

2. Go to Components > ButtonPrimary
3.

### **After**
https://github.com/user-attachments/assets/9479a519-a9d5-4b1a-a034-23404af940c9

### DIFF
--- a/packages/design-system-react-native/src/components/Button/variants/ButtonPrimary/ButtonPrimary.constants.ts
+++ b/packages/design-system-react-native/src/components/Button/variants/ButtonPrimary/ButtonPrimary.constants.ts
@@ -1,0 +1,25 @@
+import { DEFAULT_BUTTONBASE_PROPS } from '../../../../primitives/ButtonBase/ButtonBase.constants';
+import { IconColor } from '../../../Icon';
+import { TextColor } from '../../../Text';
+import type { ButtonPrimaryProps } from './ButtonPrimary.types';
+
+// Defaults
+export const DEFAULT_BUTTONPRIMARY_PROPS: Partial<ButtonPrimaryProps> = {
+  ...DEFAULT_BUTTONBASE_PROPS,
+  textProps: {
+    ...DEFAULT_BUTTONBASE_PROPS.textProps,
+    color: TextColor.PrimaryInverse,
+  },
+  spinnerProps: {
+    ...DEFAULT_BUTTONBASE_PROPS.spinnerProps,
+    color: IconColor.PrimaryInverse,
+  },
+  startIconProps: {
+    ...DEFAULT_BUTTONBASE_PROPS.startIconProps,
+    color: IconColor.PrimaryInverse,
+  },
+  endIconProps: {
+    ...DEFAULT_BUTTONBASE_PROPS.endIconProps,
+    color: IconColor.PrimaryInverse,
+  },
+};
--- a/packages/design-system-react-native/src/components/Button/variants/ButtonPrimary/ButtonPrimary.stories.tsx
+++ b/packages/design-system-react-native/src/components/Button/variants/ButtonPrimary/ButtonPrimary.stories.tsx
@@ -1,0 +1,183 @@
+import { useTailwind } from '@metamask/design-system-twrnc-preset';
+import type { Meta, StoryObj } from '@storybook/react-native';
+import { View } from 'react-native';
+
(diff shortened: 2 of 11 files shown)


packages/design-system-react-native/src/components/Button/variants/ButtonPrimary/ButtonPrimary.stories.tsx
```diff
@@ -0,0 +1,183 @@
...
+}) => {
+  const tw = useTailwind();
+  return (
+    <View
+      style={[
+        tw`${isInverse ? `bg-primary-default` : 'bg-background-default'}`,
+      ]}
+    >
+      <ButtonPrimary isInverse={isInverse} {...props} />
+    </View>
```
georgewrmarshall, line 69: Nice touch!

packages/design-system-react-native/src/primitives/ButtonBase/ButtonBase.types.ts
```diff
@@ -8,15 +8,15 @@ export enum ButtonBaseSize {
...
   /**
    * Represents a medium button size (40px).
    */
-  Md = 40,
+  Md = '40',
   /**
    * Represents a large button size (48px).
    */
-  Lg = 48,
+  Lg = '48',
```
georgewrmarshall, line 19: non-blocking: Going to write here what I mentioned in our DMs  You’re absolutely right—I forgot about reverse mapping with number enums. It’s [even mentioned in our contributor docs](https://github.com/MetaMask/contributor-docs/blob/add-changelogs-guidelines/docs/typescript.md#dont-use-numeric-enums) to avoid them :sweat_smile:

I think the most maintainable approach is to use string enums like this:

```typescript
export enum ButtonBaseSize {
  /**
   * Represents a small button size (32px).
   */
  Sm = 'Sm',
  /**
   * Represents a medium button size (40px).
   */
  Md = 'Md',
  /**
   * Represents a large button size (48px).
   */
  Lg = 'Lg',
}

// Map enum values to styles
const buttonSizes = {
  [ButtonBaseSize.Sm]: 32,
  [ButtonBaseSize.Md]: 40,
  [ButtonBaseSize.Lg]: 48,
};```

This approach has a few benefits:
1. *Cross-platform compatibility*: We can share these types between platforms.  
2. *Maintainability*: We can change how the values are mapped without affecting the enum itself.  
3. *Readability*: String enums are more descriptive and self-documenting.  

I’ll do some more research, but I think this is a solid direction. Let me know your thoughts, and I’m happy to discuss further or bring this up in #metamask-dev if we think

**Review by georgewrmarshall — DISMISSED** [review]
Looking great! Left some non-blocking comments. I noticed that this component isn't being exported from the root index which I imagine is intentional as we should only be using `Button` which might be something we should also do in design system react 🤔 
- Checked all stories in storybook ✅ 
- Checked docs ✅

File: packages/design-system-react-native/src/components/Button/variants/ButtonPrimary/ButtonPrimary.constants.ts [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+    ...DEFAULT_BUTTONBASE_PROPS.spinnerProps,
+    color: IconColor.PrimaryInverse,
+  },
+  startIconProps: {
+    ...DEFAULT_BUTTONBASE_PROPS.startIconProps,
+    color: IconColor.PrimaryInverse,
+  },
+  endIconProps: {
+    ...DEFAULT_BUTTONBASE_PROPS.endIconProps,
+    color: IconColor.PrimaryInverse,
```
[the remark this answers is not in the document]
brianacnguyen: ` ...DEFAULT_BUTTONBASE_PROPS` already took care of that

packages/design-system-react-native/src/primitives/ButtonBase/ButtonBase.types.ts
brianacnguyen: Going to keep it to what it is right now until we're sure we can safely get the values

georgewrmarshall (commented):
LGTM!

PR #449, opened 2025-02-25
[DSRN]Added AvatarBase

brianacnguyen:
## **Description**
This PR adds the `AvatarBase` component to the `design-system-react-native` package

## **Related issues**

Fixes: #357 

## **Manual testing steps**

2. Go to Primitives > AvatarBase
3.

### **After**
https://github.com/user-attachments/assets/61def6f5-26c0-4b77-955b-08233d5a144e

### DIFF
--- a/packages/design-system-react-native/src/components/TextButton/TextButton.tsx
+++ b/packages/design-system-react-native/src/components/TextButton/TextButton.tsx
@@ -135,8 +135,7 @@ const TextButton = ({
    */
 
   // Get the line height of the text, which is also the height of the component
-  const finalVariant =
-    textProps?.variant || DEFAULT_TEXTBUTTON_PROPS.textProps.variant;
+  const finalVariant = finalTextProps.variant;
   const { lineHeight } = tw`text-${finalVariant as string}`;
   const componentHeight = Number(lineHeight);
 
@@ -145,10 +144,8 @@ const TextButton = ({
   const nonTextOffSet = 2.5;
 
   // Get the final sizes of Icons to help with manual alignments
-  const finalStartIconSize =
-    startIconProps?.size || DEFAULT_TEXTBUTTON_PROPS.startIconProps.size;
-  const finalEndIconSize =
-    endIconProps?.size || DEFAULT_TEXTBUTTON_PROPS.endIconProps.size;
+  const finalStartIconSize = finalStartIconProps.size;
+  const finalEndIconSize = finalEndIconProps.size;
 
   // Offsets to vertically center the Icon
   const finalStartIconOffset =
--- a/packages/design-system-react-native/src/index.ts
+++ b/packages/design-system-react-native/src/index.ts
@@ -1,11 +1,27 @@
 import { withThemeProvider } from '@metamask/design-system-twrnc-preset';
 
-import TextComponent from './components/Text';
+import ButtonComponent from './components/Button';
+export const Button = withThemeProvider(ButtonComponent);
+export { ButtonProps, ButtonSize, ButtonVariant } from './components/Button';
 
+import ButtonIconComponent from './components/ButtonIcon';
+export const ButtonIcon = withThemeProvider(ButtonIconComponent);
+export { ButtonIconProps, ButtonIconSize } from './components/ButtonIcon';
+
+import IconComponent from './components/Icon';
+export const Icon = withThemeProvider(IconComponent);
+export { IconColor, IconName, IconProps, IconSize } from './components/Icon';
+
+import TextComponent from './components/Text';
 export const Text = withThemeProvider(TextComponent);
 export {
-  TextVariant,
-  TextColor,
   FontWeight,
   FontStyle,
+  TextColor,
+  TextProps,
+  TextVariant,
 } from './components/Text';
+
+import TextButtonComponent from './components/TextButton';
+export const TextButton = withThemeProvider(TextButtonComponent);
+export { TextButtonProps } from './components/TextButton';
--- a/packages/design-system-react-native/src/primitives/AvatarBase/AvatarBase.tsx
+++ b/packages/design-system-react-native/src/primitives/AvatarBase/AvatarBase.tsx
@@ -4,7 +4,6 @@ import React, { useMemo } from 'react';
 import { View } from 'react-native';
 
 import Text from '../../components/Text';
-import type { TextProps } from '../../components/Text/Text.types';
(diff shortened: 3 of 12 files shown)


packages/design-system-react-native/src/primitives/AvatarBase/AvatarBase.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const finalFallbackTextProps: Omit<TextProps, 'children'> = {
+    ...DEFAULT_AVATARBASE_PROPS.fallbackTextProps,
+    ...fallbackTextProps,
+  };
+
+  return (
+    <View style={[tw`${twContainerClassNames}`, style]} {...props}>
+      {fallbackText ? (
+        <Text {...finalFallbackTextProps}>{fallbackText}</Text>
```
georgewrmarshall, line 39: I suggest setting default props at the prop definition level rather than handling defaults in the component body:

```suggestion
  fallbackTextProps = DEFAULT_AVATARBASE_PROPS.fallbackTextProps,
  twClassName = '',
  style,
  ...props
}: AvatarBaseProps) => {
  const tw = useTailwind();
  const twContainerClassNames = useMemo(() => {
    return generateAvatarBaseContainerClassNames({
      size,
      shape,
      twClassName,
    });
  }, [size, shape, twClassName]);

  return (
    <View style={[tw`${twContainerClassNames}`, style]} {...props}>
      {fallbackText ? (
        <Text {...fallbackTextProps}>{fallbackText}</Text>
```

This would:
1. Eliminate unnecessary object creation on each render
2. Follow React functional component best practices
3. Make the code more concise and readable
4. Maintain consistency with how other props are handled

This pattern is already used for size, shape, and twClassName props.

packages/design-system-react-native/src/primitives/AvatarBase/README.md
```diff
@@ -0,0 +1,163 @@
...
+```tsx
+<AvatarBase shape={AvatarShape.Square}>👤</AvatarBase>
+```
+
+---
+
+### Accessibility
+
+- Ensure `fallbackText` provides meaningful information when the avatar content is unavailable.
+- Use `fallbackTextProps` to style text for better visibility and contrast.
```
georgewrmarshall, line 143: non-blocking: Do we need to note any react-native accessibility props for this component that can be passed through ...props if needed. We can also save this for the variants. Some ideas:

1. **`accessibilityLabel`**: To describe avatars that typically represent tokens, accounts, or networks.
3. **`accessibilityRole`**: Should be set to "image" by default since avatars in MetaMask function as visual identifiers for blockchain assets and accounts.
4. **`accessibilityHint`**: Particularly important for interactive avatars in MetaMask workflows: "Opens account details for this wallet"
5. **`accessible`**: Should be true by default for all avatars that represent meaningful entities in the MetaMask interface (tokens, accounts, networks), ensuring screen readers properly identify these critical UI elements.

georgewrmarshall (changes requested):
Looking good! I’ve left a few comments—let’s sync on some of these as I think it’ll help us stay aligned across platforms.  

I know we’re all keeping platform consistency in mind, and I just want to reinforce how valuable that alignment is. Keeping component names, APIs, and types consistent between React Native and React will make things easier for engineers using our components. It’ll also set us up for a smoother transition when we eventually move our types to a shared package, reducing the need for refactoring down the line.

Checked:
- Storybook stories ✅ 
- README docs ✅

packages/design-system-react-native/src/primitives/AvatarBase/AvatarBase.tsx
georgewrmarshall, line 39: I think this condition may be around the wrong way. If `children` exists we want to show the children, fallbackText should only be used if `children` don't exists?

```suggestion
      {children || <Text {...finalFallbackTextProps}>{fallbackText}</Text>}
```

### Before

https://github.com/user-attachments/assets/7a9c7063-7b50-4d6c-86e4-7cff7ff1702d

### After

https://github.com/user-attachments/assets/8985658f-651e-41a8-8429-f29b9b3eed0d

packages/design-system-react-native/src/primitives/AvatarBase/README.md
georgewrmarshall: Side note: We should also add this to the react component docs

brianacnguyen: Great suggestion. I'll add these to the variants!

packages/design-system-react-native/src/components/TextButton/TextButton.tsx
```diff
@@ -145,10 +144,8 @@ const TextButton = ({
   const nonTextOffSet = 2.5;
 
   // Get the final sizes of Icons to help with manual alignments
-  const finalStartIconSize =
-    startIconProps?.size || DEFAULT_TEXTBUTTON_PROPS.startIconProps.size;
-  const finalEndIconSize =
-    endIconProps?.size || DEFAULT_TEXTBUTTON_PROPS.endIconProps.size;
+  const finalStartIconSize = finalStartIconProps.size;
+  const finalEndIconSize = finalEndIconProps.size;
```
georgewrmarshall, line 149: non-blocking: Not sure if this was intended on being included but I think it's fine

georgewrmarshall (approved):
LGTM! I still think we should consider using the React approach for setting default props to avoid creating extra objects on render, but I’ll do some more research on the pros and cons of this approach.

packages/design-system-react-native/src/index.ts
```diff
@@ -1,11 +1,27 @@
...
   FontWeight,
   FontStyle,
+  TextColor,
+  TextProps,
+  TextVariant,
 } from './components/Text';
+
+import TextButtonComponent from './components/TextButton';
+export const TextButton = withThemeProvider(TextButtonComponent);
+export { TextButtonProps } from './components/TextButton';
```
brianacnguyen, line 27: @georgewrmarshall I was debating whether to export non-official components (AvatarBase, ButtonBase, TextOrChildren, ButtonPrimary, ButtonSecondary, ButtonTertiary, ButtonAnimated). Let me know your thoughts on these

georgewrmarshall (approved):

packages/design-system-react-native/src/index.ts
georgewrmarshall: Good question! I think base components are worth exporting, but any variants with a unification component probably aren't. What do you think?

PR #555, opened 2025-04-16
[DSR] Added AvatarGroup

brianacnguyen:
## **Description**
This PR adds the `AvatarGroup` component to the `@metamask/design-system-react` package

## **Related issues**

Fixes: #372 

## **Manual testing steps**

1. Run `yarn storybook`
2. Go to React Components > AvatarGroup
3.

### **After**
https://github.com/user-attachments/assets/59a4a7fe-9dbf-41d1-a92a-242ef230cb2c


apps/storybook-react/tailwind.config.js
```diff
@@ -9,4 +9,5 @@ module.exports = {
     './src/**/*.{js,jsx,ts,tsx}',
     './stories/**/*.{js,jsx,ts,tsx}',
   ],
+  mode: 'jit',
```
georgewrmarshall, line 12: Could this be in it's own PR?

packages/design-system-react/src/components/AvatarGroup/README.md
```diff
@@ -0,0 +1,262 @@
+# AvatarGroup
```
georgewrmarshall, line 1: Should this be an `mdx` file?

packages/design-system-react/src/components/index.ts
```diff
@@ -11,6 +11,13 @@ export type { AvatarBaseProps } from './AvatarBase';
 export { AvatarFavicon, AvatarFaviconSize } from './AvatarFavicon';
 export type { AvatarFaviconProps } from './AvatarFavicon';
 
+export {
+  AvatarGroup,
+  AvatarGroupSize,
+  AvatarGroupVariant,
+} from './AvatarGroup';
+export type { AvatarGroupProps } from './AvatarGroup';
```
georgewrmarshall, line 19: Should we be exporting these from the root `index.ts` as well?

apps/storybook-react/tailwind.config.js
georgewrmarshall, line 12: I believe JIT mode is the [default in v3](https://tailwindcss.com/blog/tailwindcss-v3#just-in-time-all-the-time) and above so this config option is deprecated and doesn't do anything. Have you read something different?

packages/design-system-react/__mocks__/svgMock.js
```diff
@@ -0,0 +1 @@
+module.exports = 'test-file-stub.svg';
```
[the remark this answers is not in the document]
brianacnguyen: Apologies. This is to help jest out. Jest tries to interpret the locally imported svgs in the constants as JavaScript — and sees an unexpected <svg> tag. This will help Jest will stop crashing when running tests on files that import .svg images

packages/design-system-react/src/components/index.ts
brianacnguyen: Screenshot 2025-04-17 at 10 27 04 AM

georgewrmarshall (commented):
This is looking great! 🔥 I've left one suggestion regarding the avatar size adjustments—might've missed it in the DSRN version, apologies! Also noticed that sample data constants are currently bundled with production code. We can avoid this by moving them to a separate file. Let's sync on the best approach and handle that in a follow-up PR.

packages/design-system-react/src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+                <AvatarAccount
+                  hasBorder
+                  key={`avatar-${index}`}
+                  {...(avatarProps as AvatarAccountProps)}
+                  size={size as unknown as AvatarBaseSize}
+                  className={`${index > 0 && negativeSpaceBetweenAvatarsTwClassName} ${avatarProps.className}`}
+                  style={{
+                    zIndex: index + 1,
+                    ...avatarProps.style,
+                  }}
```
georgewrmarshall, line 83: We generally want to avoid using z-index as it can cause complications with the layer order. I think we should be able to achieve the correct layering including the overlay without z-index

packages/design-system-react/src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+                  data-testid={`avatar-${variant}`}
```
georgewrmarshall, line 84: Typically we don't want data-testids set in the component code. They should be applied in the unit tests using the avatarProps props object

packages/design-system-react/__mocks__/svgMock.js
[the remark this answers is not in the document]
georgewrmarshall: I think this exposes a larger issue with the locally imported assets? We may want to revert to using urls for now even if it's less than ideal.

packages/design-system-react/src/components/AvatarGroup/README.mdx
```diff
@@ -0,0 +1,300 @@
...
+    },
+    {
+      src: { uri: 'https://example.com/token2.png' },
+    },
+    {
+      src: { uri: 'https://example.com/token3.png' },
+    },
+  ]}
+/>;
+```
```
georgewrmarshall, line 78: Another reason to treat stories as self-documenting code is that we can rely on the *Show code* feature instead of needing to provide separate usage examples. Here are a few examples that show how this varies depending on how we write our stories—apologies for not flagging this earlier!

- The first example isn’t very helpful as a usage reference.

Screenshot 2025-04-18 at 12 37 58 PM

- The second is better but still hides the types behind constants, which makes it harder to understand at a glance.

Screenshot 2025-04-18 at 12 39 16 PM

- The third example is ideal—it uses the exact types and values an engineer would need to implement the button, making it much more usable.
- 
Screenshot 2025-04-18 at 12 39 08 PM

brianacnguyen: so how would you write these stories? The story you pointed was part of the default?
Screenshot 2025-04-22 at 12 41 53 PM

brianacnguyen: so should we omit these parts? 
Screenshot 2025-04-22 at 12 45 38 PM

brianacnguyen: IMO the stories that we write, especially we code samples CLEARLY displayed in every section is good enough. If we also update the actual code section to have a flat map like this, it'll make the stories file large and unorganized

packages/design-system-react/src/components/AvatarGroup/AvatarGroup.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+    className: '',
+  },
+  render: (args: Omit<AvatarGroupProps, 'avatarPropsArr'>) => (
+    <AvatarGroupStory {...args} />
+  ),
+};
+
+export const Variant: Story = {
+  render: () => (
+    <div style={{ display: 'flex', flexDirection: 'column', gap: 16 }}>
```
georgewrmarshall, line 111: Can we use tailwind classnames instead of inline styles

georgewrmarshall (commented):
Looking good. Did you find out why we need the settings in the svg and can we use tailwind classnames as opposed to inline styles in the stories

packages/design-system-react/jest.config.js
```diff
@@ -35,6 +36,9 @@ module.exports = merge(baseConfig, {
       statements: 100,
     },
   },
+  moduleNameMapper: {
+    '\\.svg$': '<rootDir>/__mocks__/svgMock.js',
+  },
```
georgewrmarshall, line 41: Can you provide some context on this change and why it is needed?

packages/design-system-react/src/components/AvatarGroup/AvatarGroup.dev.ts
```diff
@@ -0,0 +1,131 @@
...
+    src: SAMPLE_AVATARFAVICON_URIS[4],
+  },
+  {
+    name: 'Slack',
+    src: SAMPLE_AVATARFAVICON_URIS[5],
+  },
+  {
+    name: 'Zoom',
+    src: SAMPLE_AVATARFAVICON_URIS[6],
+  },
```
georgewrmarshall, line 68: Can we use the same assets as Figma? So we have more alignment https://www.figma.com/design/HKpPKij9V3TpsyMV1TpV7C/%F0%9F%A7%ACDS-Components?m=auto&node-id=12890-38062
Screenshot 2025-04-23 at 8 46 11 AM

brianacnguyen: Figma does not have 7 assets of each variant

packages/design-system-react/jest.config.js
brianacnguyen: Removed

packages/design-system-react/jest.config.js
```diff
@@ -21,6 +21,7 @@ module.exports = merge(baseConfig, {
     'src/components/Icon/icons/', // Auto-generated icons
     'scripts/create-component/ComponentName/', // component template
     '\\.constants\\.(ts|js)$', // ignore all .constants.ts and .constants.js files
+    '\\.dev\\.(ts|js)$', // ignore all .constants.ts and .constants.js files
```
georgewrmarshall, line 24: Would it make sense to ignore these files from the `tsconfig.ts` so they aren't included in the build? I'm not sure if it matters if they're ignored from coverage or not they could

georgewrmarshall (approved):

PR #559, opened 2025-04-17
[DSR] Added BadgeNetwork

brianacnguyen:
## **Description**
This PR adds the `BadgeNetwork` component to the `@metamask/design-system-react`

## **Related issues**

Fixes: #395 

## **Manual testing steps**

1. Run `yarn storybook`
2. Go to React Components > BadgeNetwork
3.

### **After**
https://github.com/user-attachments/assets/7b9cc9b0-e621-4303-9f18-371eb780a3cb


packages/design-system-react/src/components/BadgeNetwork/BadgeNetwork.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+  it('always applies Xs size + border classes for the outer element', () => {
+    render(<BadgeNetwork name="TestNet" src="icon.png" data-testid="badge" />);
+
+    const wrapper = screen.getByTestId('badge');
+    const classList = wrapper.className.split(/\s+/);
+
+    expect(classList).toEqual(
+      expect.arrayContaining([
+        'h-[17px]',
+        'w-[17px]',
```
georgewrmarshall, line 16: Once we are using the correct sizes for Avatars with border this test will need to be updated we may not necessarily want to test for default props

packages/design-system-react/src/components/BadgeNetwork/BadgeNetwork.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import { AvatarNetwork, AvatarNetworkSize } from '../AvatarNetwork';
+import type { BadgeNetworkProps } from './BadgeNetwork.types';
+
+export const BadgeNetwork = React.forwardRef<HTMLDivElement, BadgeNetworkProps>(
+  ({ ...props }, ref) => {
+    return <AvatarNetwork {...props} size={AvatarNetworkSize.Xs} hasBorder />;
```
georgewrmarshall, line 8: non-blocking: are we certain we want to enforce this on the code level? It will mean we will have to refactor this component if a size is needed.

georgewrmarshall (approved):
Nice work 💯 LGTM!

packages/design-system-react/src/components/BadgeNetwork/BadgeNetwork.test.tsx
brianacnguyen: default props is still a test we should check. if a user does not enter a prop and theres a default to it, why dont you think its necessary to check if the default is rendered correctly?

brianacnguyen: and this is the correct size for Avatars with border (16px avatar + 1 px border)

packages/design-system-react/src/components/BadgeNetwork/BadgeNetwork.tsx
brianacnguyen: From what I remember this was a design call to lock it

georgewrmarshall (approved):

packages/design-system-react/src/components/BadgeNetwork/BadgeNetwork.tsx
georgewrmarshall: Thanks, Brian. I totally get that this was a design decision, but do you think they would have made the same call with a better understanding of the technical impact—like if this does change, needing to update the component, rewrite tests, create a release, and import it across platforms?

Our usual convention is to set sensible defaults while allowing flexibility through prop spreading. It gives design control without locking us into changes that require significant rework. Just wondering if that context was fully clear when the decision was made.

georgewrmarshall (approved):

Issue #3, opened 2024-09-20
Host storybook-react for metamask-design-system on GitHub Pages with Auto-Update on Merges to main


georgewrmarshall:
### **Description**

We need to host `storybook-react` for the `metamask-design-system` React component library on a GitHub page. This Storybook should be automatically updated whenever changes are merged into the `main` branch to ensure that the documentation reflects the latest updates.

### **Technical Details**

- Configure the GitHub repository to deploy Storybook to GitHub Pages.
- Set up GitHub Actions or a similar CI/CD pipeline to automatically build and publish Storybook on any merge to the `main` branch.
- Ensure that the deployment happens smoothly and the hosted Storybook reflects all the latest updates from the `metamask-design-system` repository.
- Verify that the Storybook documentation is accessible through a consistent, public URL (e.g., via `https://metamask.github.io/metamask-design-system/`).

### **Acceptance Criteria**

- Storybook for the `metamask-design-system` is hosted on GitHub Pages.
- The hosted Storybook is automatically updated whenever changes are merged to the `main` branch.
- The deployment process is stable and ensures no downtime during updates.
- Stakeholders can review the deployed Storybook without needing local setup.

### **References**
- Design tokens has the same implementation we want to achieve https://github.com/MetaMask/design-tokens

PR #610, opened 2025-04-27
chore: adding github workflow to publish storybook react

georgewrmarshall:
## **Description**

This PR adds GitHub Actions workflow to automatically build and deploy the Storybook React to GitHub Pages whenever changes are merged to the `main` branch.
 - Triggers on pushes to `main` branch
 - Uses the official github actions [actions/upload-pages-artifact@v3](https://github.com/actions/upload-pages-artifact?) and  [actions/deploy-pages@v4](https://github.com/actions/deploy-pages?tab=readme-ov-file) as suggested in [this comment](https://github.com/MetaMask/metamask-design-system/pull/610#discussion_r2069333465)
 - Doesn't require a secret because official actions use GitHub's built-in [OpenID Connect (OIDC)](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect) authentication.

## **Related issues**

Fixes: #3

## **Manual testing steps**
The action has been successfully tested(again) in a fork environment. You can see the working deployment at [georgewrmarshall.github.io/metamask-design-system](https://georgewrmarshall.github.io/metamask-design-system/?path=/docs/getting-started-introduction--docs)

1. Merge this PR to main branch
2. Wait for the GitHub Action to complete (~2-3 minutes)
3. Visit https://metamask.github.io/metamask-design-system/ to verify Storybook is deployed

### **Before**
N/A - No previous Storybook deployment

### **After**
- Successfully deployed Storybook: [Fork Demo](https://georgewrmarshall.github.io/metamask-design-system/)
- GitHub Action running successfully in fork environment

https://github.com/user-attachments/assets/021669e7-4708-45b5-a18d-ef74775c4a12

## **Technical Implementation Details**
- Workflow uses workspace-aware commands to build dependencies
- Correctly handles monorepo structure with appropriate working directory settings
- Uses relative paths for GitHub Pages deployment to ensure compatibility
- Maintains proper build order: tokens → tailwind preset → react components → storybook

- [x] I've applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-extension/blob/develop/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.


.github/workflows/storybook-deploy.yml
```diff
@@ -0,0 +1,40 @@
+name: Build and Deploy Storybook
```
georgewrmarshall, line 1: Copied from the archived [design tokens repo](https://github.com/MetaMask/design-tokens/blob/main/.github/workflows/build-deploy-storybook.yml)

.github/workflows/storybook-deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  workflow_call:
+    secrets:
+      PUBLISH_STORYBOOK_TOKEN:
```
georgewrmarshall, line 9: This needs to be created will reach out to appropriate folks

.github/workflows/storybook-deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+          cache-node-modules: true
+
+      - name: Install dependencies
+        run: yarn --immutable
+
+      - name: Build Storybook
+        run: yarn build-storybook
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@de7ea6f8efb354206b205ef54722213d99067935
```
georgewrmarshall, line 36: This action is used by a [few MM repos](https://github.com/search?q=org%3AMetaMask%20peaceiris%2Factions-gh-pages%40de7ea6f8efb354206b205ef54722213d99067935&type=code) to publish to github pages

.github/workflows/storybook-deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+    environment: github-pages
+    defaults:
+      run:
+        working-directory: apps/storybook-react
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v1
+        with:
+          is-high-risk-environment: false
+          cache-node-modules: true
```
georgewrmarshall, line 27: Not sure if there is an alternative here to using `cache-node-modules`. Maybe @itsyoboieltr knows?

georgewrmarshall:
After further inspection I'm not sure this action will work the same as it did in the design tokens repo. Will try testing out the PR version in this PR first https://github.com/MetaMask/metamask-design-system/pull/613

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,45 @@
+name: Build Storybook Main
```
georgewrmarshall, line 1: Copied from the archived [design tokens repo](https://github.com/MetaMask/design-tokens/blob/main/.github/workflows/build-deploy-storybook.yml) with some updates for monorepo structure

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,45 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  workflow_call:
+    secrets:
+      PUBLISH_STORYBOOK_TOKEN:
```
georgewrmarshall, line 9: This needs to be created. I will reach out to appropriate folks.

georgewrmarshall: @mcmire would you be able to create this `PUBLISH_STORYBOOK_TOKEN ` token or point it to this repo if it is still created for when it was used for the [design tokens repo](https://github.com/MetaMask/design-tokens/settings/secrets/actions) please?

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          yarn workspace @metamask/design-tokens build
+          yarn workspace @metamask/design-system-tailwind-preset build
+          yarn workspace @metamask/design-system-react build
+
+      - name: Build Storybook
+        run: yarn build-storybook
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@de7ea6f8efb354206b205ef54722213d99067935
```
georgewrmarshall, line 40: This action is used by a [few MM repos](https://github.com/search?q=org%3AMetaMask%20peaceiris%2Factions-gh-pages%40de7ea6f8efb354206b205ef54722213d99067935&type=code) to publish to github pages but open to using something else if it exists

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          personal_token: ${{ secrets.PUBLISH_STORYBOOK_TOKEN }}
+          publish_dir: apps/storybook-react/storybook-static
+          destination_dir: .
+          force_orphan: true
```
georgewrmarshall, line 45: The f`orce_orphan: true` setting tells Git to create a new history for the `gh-pages` branch, completely disconnected from the main repository's history. Here's what it does specifically:
- Creates a fresh gh-pages branch with only one commit
- Discards all previous history of the gh-pages branch
- Each deployment creates a new, clean branch with just the current build files

Benefits of using `force_orphan: true:`
- Keeps the gh-pages branch clean and lightweight
- Prevents the branch from accumulating deployment history
- Reduces repository size since old deployment histories are removed
- Avoids merge conflicts between deployments
Without force_orphan: true, the action would:
Keep all previous deployment history
Add new commits on top of the existing branch
Potentially grow the repository size over time with deployment history
This is particularly useful for deployment branches like gh-pages where we only care about the current state of the built files, not their history.

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          is-high-risk-environment: false
+          cache-node-modules: true
+
+      - name: Build required packages
+        working-directory: .
+        run: |
+          yarn workspace @metamask/design-tokens build
+          yarn workspace @metamask/design-system-tailwind-preset build
+          yarn workspace @metamask/design-system-react build
```
georgewrmarshall, line 34: This is specific to the monorepo and different from the original workflow because in a single repository, all these would typically be one package. In a monorepo, we need to explicitly build each package in the correct order to ensure all dependencies are available and up-to-date before building Storybook.

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,45 @@
...
+    environment: github-pages
+    defaults:
+      run:
+        working-directory: apps/storybook-react
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v1
+        with:
+          is-high-risk-environment: false
+          cache-node-modules: true
```
georgewrmarshall, line 27: Not sure if there is an alternative here to using `cache-node-modules` I've seen in slack discussions that this might be unsafe. Maybe @itsyoboieltr knows?

.github/workflows/build-storybook-main.yml
mcmire: This has been added!

.github/workflows/build-storybook-main.yml
mcmire: I believe we commonly use this action in other repos because the official action doesn't support deploying to a subdirectory. This is common for libraries because we want to have a recorded version of API docs for each historical version, so we deploy to e.g. `1.2.3/` within `gh-pages` instead of the root directory. In this case however we seem to be using the action in a different way so perhaps using an unofficial action isn't necessary.

.github/workflows/build-storybook-main.yml
georgewrmarshall: Thank you!

.github/workflows/build-storybook-main.yml
georgewrmarshall: Great point! I could have been clearer in my earlier comment — this Storybook build currently uses only these specific packages, so I thought there might be an opportunity to save some time by including only the relevant ones. However, `yarn build` is probably more sustainable incase dependencies change. I will update to `yarn build`

.github/workflows/build-storybook-main.yml
georgewrmarshall: Thanks for the suggestion and context both! I'll update with the official action here

.github/workflows/build-storybook-main.yml
georgewrmarshall: Thanks for the detailed explanation and suggestion. I've updated here https://github.com/MetaMask/metamask-design-system/pull/610/commits/024a16f1a502a02ff7a93bc9f3f51c93beb6631d

.github/workflows/build-storybook-main.yml
georgewrmarshall: Updated to align with [actions/upload-pages-artifact@v3](https://github.com/actions/upload-pages-artifact?) and  [actions/deploy-pages@v4](https://github.com/actions/deploy-pages?tab=readme-ov-file) here https://github.com/MetaMask/metamask-design-system/pull/610/commits/4cc94306a023d83cf05cb0bec676628e94fd7db8

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Build Storybook
+        run: yarn build-storybook
+
+      - name: Upload artifact
+        id: upload
+        uses: actions/upload-pages-artifact@v3
+        with:
+          name: github-pages
+          path: apps/storybook-react/storybook-static
+          retention-days: 1
```
georgewrmarshall, line 40: Uses the official github action example [upload-pages-artifact](https://github.com/actions/upload-pages-artifact?tab=readme-ov-file)

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,55 @@
...
+      pages: write
+      id-token: write
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
```
georgewrmarshall, line 55: Uses the official gitHub action example [actions/deploy-pages@v4](https://github.com/actions/deploy-pages)

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,55 @@
+name: Build Storybook Main
+
+on:
+  push:
+    branches:
+      - main
+  workflow_call:
+    secrets:
+      PUBLISH_STORYBOOK_TOKEN:
+        required: true
```
mcmire, line 10: It looks like this token is no longer being used because the `deploy-pages` action mints its own token. So maybe we can remove the secret requirement?

``` suggestion
```

.github/workflows/build-storybook-main.yml
```diff
@@ -0,0 +1,51 @@
+name: Build Storybook Main
+
+on:
+  push:
+    branches:
+      - main
```
georgewrmarshall, line 6: The workflow no longer uses the third-party `peaceiris/actions-gh-pages` action which required a personal access token for authentication. e.g

```yml
// before
`PUBLISH_STORYBOOK_TOKEN`
```

Instead, we've migrated to GitHub's official Pages deployment actions (`upload-pages-artifact` and `deploy-pages`). These official actions use GitHub's built-in OpenID Connect (OIDC) authentication system with fine-grained permissions:

```yaml
permissions:
  pages: write
  id-token: write
```

This approach:
- Eliminates the need to create, store, and rotate a personal access token
- Follows security best practices with short-lived tokens generated during workflow runs
- Uses the principle of least privilege with fine-grained permissions
- Aligns with GitHub's recommended deployment pattern for Pages

.github/workflows/build-storybook-main.yml
georgewrmarshall: Yes! Thanks Elliot, @itsyoboieltr also pointed out I'm no longer using the secret and after further investigation it's not longer needed! [See comment](https://github.com/MetaMask/metamask-design-system/pull/610#discussion_r2072231028)

brianacnguyen (approved):

PR #736, opened 2025-06-10
[DSR] Updated Buttons to align with Brand Evolution

brianacnguyen:
## **Description**
This PR updated the `ButtonBase`, `Button`, `ButtonPrimary`, `ButtonSecondary`, `ButtonTertiary` components in the `@metamask/design-system-react` package 
- Updated design to align with Brand Evolution
- Added accessibility
- Updated Mapping names 
- Added focus states

## **Related issues**

Fixes: #735 

## **Manual testing steps**

1. Run `yarn storybook`
2. Check `ButtonBase`, `Button`, `ButtonPrimary`, `ButtonSecondary`, `ButtonTertiary`
3.

### **After**
ButtonBase
https://github.com/user-attachments/assets/ee2fdd7e-d293-4106-9178-461a6b1b8ca4

ButtonPrimary
https://github.com/user-attachments/assets/bea01d78-af80-4ef6-95d5-ca6c9db31623

ButtonSecondary
https://github.com/user-attachments/assets/8105708f-d7f7-4d51-8143-612cf4a0c9e5

ButtonTertiary
https://github.com/user-attachments/assets/fdd74fc6-858c-4ba5-b3a2-ce337788ca0b

### DIFF
--- a/packages/design-system-react/src/components/Button/Button.test.tsx
+++ b/packages/design-system-react/src/components/Button/Button.test.tsx
@@ -236,6 +236,151 @@ describe('Button', () => {
     });
   });
 
+  describe('Focus Styling', () => {
+    describe('Primary Button Focus', () => {
+      it('applies correct focus outline for non-inverse primary button', () => {
+        render(<Button>Primary Button</Button>);
+        const button = screen.getByRole('button');
+        expect(button).toHaveClass(
+          'focus-visible:outline-none',
+          'focus-visible:ring-0',
+          'focus-visible:outline',
+          'focus-visible:outline-2',
+          'focus-visible:outline-primary-default',
+          'focus-visible:outline-offset-2',
+        );
+      });
+
+      it('applies correct focus outline for inverse primary button', () => {
+        render(<Button isInverse>Primary Inverse Button</Button>);
+        const button = screen.getByRole('button');
+        expect(button).toHaveClass(
+          'focus-visible:outline-none',
+          'focus-visible:ring-0',
+          'focus-visible:outline',
+          'focus-visible:outline-2',
+          'focus-visible:outline-background-default',
+          'focus-visible:outline-offset-4',
+        );
+      });
+
+      it('applies correct focus outline for danger primary button', () => {
+        render(<Button isDanger>Primary Danger Button</Button>);
+        const button = screen.getByRole('button');
+        expect(button).toHaveClass(
+          'focus-visible:outline-none',
+          'focus-visible:ring-0',
+          'focus-visible:outline',
+          'focus-visible:outline-2',
+          'focus-visible:outline-primary-default',
+          'focus-visible:outline-offset-2',
+        );
+      });
+    });
+
+    describe('Secondary Button Focus', () => {
+      it('applies correct focus outline for non-inverse secondary button', () => {
+        render(
+          <Button variant={ButtonVariant.Secondary}>Secondary Button</Button>,
+        );
+        const button = screen.getByRole('button');
+        expect(button).toHaveClass(
+          'focus-visible:outline-none',
+          'focus-visible:ring-0',
+          'focus-visible:outline',
+          'focus-visible:outline-2',
+          'focus-visible:outline-primary-default',
+          'focus-visible:outline-offset-2',
+        );
+      });
+
+      it('applies correct focus outline for inverse secondary button', () => {
+        render(
+          <Button variant={ButtonVariant.Secondary} isInverse>
+            Secondary Inverse Button
+          </Button>,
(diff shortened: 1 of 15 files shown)


packages/design-system-react/src/components/Button/Button.test.tsx
```diff
@@ -155,7 +155,7 @@ describe('Button', () => {
 
       let button = screen.getByRole('button');
       expect(button).toBeDisabled();
-      expect(screen.getByText('Loading...')).toBeInTheDocument();
+      expect(screen.getAllByText('Loading...')).toHaveLength(2); // Both visible and screen reader text
```
georgewrmarshall, line 158: In this test, we check that 'Loading...' appears twice—presumably once as visible text and once for screen readers. On the web, aren’t screen readers already able to announce the visible button text without needing a separate label? Does this implementation create redundancy for screen reader users (causing 'Loading...' to be announced twice), or is there a specific accessibility reason for having both? Would it be more accessible to have only one instance, or is there a case where two are needed? I’m considering testing this to see what actually happens.

georgewrmarshall, line 158: non-blocking: In this test, we check that 'Loading...' appears twice—presumably once as visible text and once for screen readers. On the web, aren’t screen readers already able to announce the visible button text without needing a separate label? Does this implementation create redundancy for screen reader users (causing 'Loading...' to be announced twice), or is there a specific accessibility reason for having both? Would it be more accessible to have only one instance, or is there a case where two are needed? I’m considering testing this to see what actually happens.

packages/design-system-react/src/components/Button/Button.test.tsx
```diff
@@ -236,6 +236,142 @@ describe('Button', () => {
...
+        expect(button).toHaveClass(
+          'focus-visible:ring-0',
+          'focus-visible:outline',
+          'focus-visible:outline-2',
+          'focus-visible:outline-primary-default',
+          'focus-visible:outline-offset-2',
+        );
+      });
+    });
+  });
```
georgewrmarshall, line 373: non-blocking: I think these focus states look great but I'm wondering if there is a way we can set it globally so there is consistency across all components 🤔 


https://github.com/user-attachments/assets/c6611da7-e09f-427b-a757-bb6a7d6058c1

packages/design-system-react/src/components/ButtonBase/ButtonBase.tsx
```diff
@@ -151,6 +180,17 @@ export const ButtonBase = forwardRef<HTMLButtonElement, ButtonBaseProps>(
...
+        aria-busy={isLoading ? 'true' : undefined}
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabelledby}
+        aria-describedby={ariaDescribedby}
+        aria-pressed={ariaPressed}
+        aria-expanded={ariaExpanded}
+        aria-controls={ariaControls}
+        aria-haspopup={ariaHaspopup}
+        role={asChild ? undefined : 'button'}
+        tabIndex={getTabIndex()}
```
georgewrmarshall, line 193: For native <button> elements, it's not necessary to set tabIndex={0} because they are already focusable and included in the tab order by default. Setting tabIndex={0} explicitly is redundant unless you have a specific reason, such as restoring focusability after it was removed. Typically, we only need to set tabIndex for custom interactive elements that don’t have native keyboard support. I tested with and without and the behavior is the same.


https://github.com/user-attachments/assets/c360c91e-145d-40ee-aaf7-76bf8017cd92

```suggestion
```

georgewrmarshall (commented):
Great updates! I initially thought there might be redundancy for screen reader users if both the label and children mention that text is loading, but I tested it with VoiceOver and it worked as expected — nice job.

One suggestion: consider removing the `tabIndex`, as it seems somewhat redundant in this context.

https://github.com/user-attachments/assets/74c4b137-cd8a-49e2-8775-05144979e51f

Also, for future reference, I think this could have been split into two PRs — one for the button color updates and another for the accessibility improvements. 🤙

packages/design-system-react/src/components/Button/Button.test.tsx
brianacnguyen: They don’t announce the same thing twice—screen readers skip the visible “Loading…” (it’s hidden with aria-hidden="true") and only read the live-region version when loading starts. If you dropped the live region, screen reader users wouldn’t get immediate feedback

georgewrmarshall: Right! Also tested and it works well. Sorry should have updated this comment

PR #920, opened 2026-02-18
feat: add corporate-fare icon

georgewrmarshall:
## **Description**

Added CorporateFare icon to both React and React Native packages.

This PR adds the corporate-fare icon (building/office) to support use cases where a corporate or business entity representation is needed in the UI.

## **Related issues**

N/A

## **Manual testing steps**

1. Start React Storybook: `yarn storybook`
2. Navigate to Icon component stories
3. Select CorporateFare icon from controls
4. Verify icon renders correctly with different sizes and colors
5. Start React Native Storybook: `yarn storybook:ios`
6. Navigate to Icon component stories
7. Select CorporateFare icon
8. Verify icon renders correctly
9. Test on Android: `yarn storybook:android`
10. Verify icon renders correctly on Android

### **Before**

Corporate fare icon was not available.

### **After**

Corporate fare icon is now available in both React and React Native packages.

Screenshot 2026-02-18 at 8 49 52 AM
Screenshot 2026-02-18 at 8 51 40 AM


packages/design-system-react-native/src/components/Icon/assets/corporate-fare.svg
kirillzyusko: In mobile development (and it's applicable for web too I believe) it's better to create a custom font with icons.

It will be faster for parsing/rendering, especially when you render icons inside list etc. because when you render a an icon as a glyph you literally need to render only a single character in text (which is super fast)

But when you render SVG you need to parse it, pass it to native side (sometimes can be a huge string), make rendering and it's much more expensive operation.

I'm not saying that we need to use new approach for rendering icons in this PR, but it may be something that we should consider for the future? 👀 

This approach works only for monochrome icons (i. e. that have only single color). For multi-color icons such trick will not work.

Not sure, maybe we already discovered this possibility and decided to stick with .svg approach - let me know if I missed a decision document for that!

PR #1052, opened 2026-04-08
feat: [DSRN] Added TitleHub

brianacnguyen:
## **Description**

This PR adds **TitleHub** to `@metamask/design-system-react-native`: a layout component for a **title row**, an optional **amount** line with optional **amountAccessory**, an optional **bottom label** row with **bottomLabelAccessory**, and an optional **bottomAccessory** when the bottom label row is not used.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-674

## **Manual testing steps**

1. From the repo root, run React Native Storybook (e.g. `yarn storybook:ios` or `yarn storybook:android` per your setup).
2. Open **Components → TitleHub** and review **Default**, **Amount**, **Title / TitleAccessory**, **BottomLabel / BottomLabelAccessory**, **BottomAccessory**, and **AmountAccessory**.
3. Confirm layout and typography match expectations on light/dark backgrounds if applicable.

### **After**

https://github.com/user-attachments/assets/e457a1cd-a45b-4793-b38c-de102d5cbaad

## **Pre-merge author checklist**

- [x] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs)
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-extension/blob/develop/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

---

> [!NOTE]
> **Low Risk**
> Primarily additive UI component work with new exports, stories, and tests; low risk aside from potential minor layout/typography regressions where newly adopted.
> 
> **Overview**
> Introduces new `TitleHub` layout component to `@metamask/design-system-react-native`, rendering a stacked header with a required `title` row plus optional `amount`, `bottomLabel`, and a mutually exclusive `bottomAccessory`, each supporting inline end accessories.
> 
[...]

PR #1051, opened 2026-04-08
feat: [DSRN ]Added TitleStandard

brianacnguyen:
## **Description**

Adds **`TitleStandard`** to `@metamask/design-system-react-native`: a left-aligned title block with optional **top row** (`topAccessory`), **inline title accessory** (`titleAccessory`), and a **bottom row** that is either **`bottomLabel`** (with default secondary typography when it’s a string) or **`bottomAccessory`** when there is no renderable `bottomLabel`.

Shared layout props live in **`TitleStandardPropsShared`** in `@metamask/design-system-shared` (ADR-0004 alignment with other shared types). The RN implementation uses **`BoxHorizontal`** / **`TextOrChildren`** patterns, `titleProps` / `bottomLabelProps` for platform `Text`, and `twClassName` + `View` props on the root.

Also includes **unit tests**, a **README**, **Storybook** stories (and registration in `storybook.requires.js`), and a **MIGRATION.md** section for **0.14.0 → 0.15.0** documenting the stable API: **`title` is required**, **`topLabel` / `topLabelProps` removed** in favor of **`topAccessory`**, and **bottom row** behavior when both `bottomLabel` and `bottomAccessory` are passed.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-582

## **Manual testing steps**

1. From the repo root, run React Native Storybook (`yarn storybook:ios` or `yarn storybook:android` per your setup).
2. Open **Components → TitleStandard** and confirm **Default**, **Title**, **TitleAccessory**, **TopAccessory**, **BottomLabel**, and **BottomAccessory** match expected layout and copy.
3. Run package tests: `yarn workspace @metamask/design-system-react-native run jest -- TitleStandard` (or your usual Jest invocation for that workspace).

## **Screenshots/Recordings**

_If applicable, add screenshots and/or recordings to visualize the before and after of your change._

### **Before**

_N/A — new component._

### **After**

https://github.com/user-attachments/assets/3e7ae0a5-e99d-4932-91ee-c0b757f39e4b

## **Pre-merge author checklist**

- [x] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs)
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
[...]

PR #1059, opened 2026-04-09
feat: [DSRN] Add TitleSubpage

brianacnguyen:
## **Description**

Adds **`TitleSubpage`** to `@metamask/design-system-react-native`: a layout for subpage headers with a required **identity row** (leading `titleAvatar` in a 40×40 centered slot + title stack), optional **subtitle**, **amount** (display typography when a string), optional **bottom label** or **bottom accessory**, and **end accessories** on title, subtitle, amount, and bottom label rows.

---

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-583

---

## **Manual testing steps**

1. From the repo root, run React Native Storybook (e.g. `yarn storybook:ios` or `yarn storybook:android`).
2. Open **Components → TitleSubpage** and review **Default**, **Amount**, accessory variants, **BottomLabel** / **BottomAccessory**, and controls for text fields where enabled.
3. Confirm layout: avatar slot, title/subtitle/amount stacking, and bottom row behavior matches expectations.

---

## **Screenshots/Recordings**

_If applicable, add screenshots and/or recordings to visualize the before and after of your change._

### **Before**

_N/A — new component (or attach Storybook before/after if replacing something in-app)._

### **After**

https://github.com/user-attachments/assets/3ac61e0e-86b9-48d4-a167-bb1e5ed86547

---

---

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are additive (new component, shared types, stories, and tests) with no modifications to existing runtime behavior beyond new exports.
> 
> **Overview**
> Adds a new React Native `TitleSubpage` component for subpage header layouts, supporting a required identity row (`titleAvatar` + title stack) plus optional `subtitle`, `amount`, and either a `bottomLabel` row or a custom `bottomAccessory`, each with optional end-accessories.
> 
> Exports the component from `design-system-react-native`, introduces the corresponding shared props type (`TitleSubpagePropsShared`) in `design-system-shared`, and includes Storybook stories, unit tests, and a detailed README documenting composition and prop behavior (notably passing `titleAvatar` directly as the identity `BoxRow` `startAccessory`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c5a1026d1ed4c618f81a4d63d6d55ae3d866c02. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+    <TitleSubpage
+      titleAvatar={<StoryTitleAvatar />}
+      title={USDC_TITLE}
+      subtitle={USDC_SUBTITLE}
+      amount={USDC_AMOUNT}
+      amountEndAccessory={
+        <Box twClassName="ml-2">
+          <Icon
+            name={IconName.Info}
+            size={IconSize.Sm}
```
amandaye0h: Image

Let's do medium, sm is too small

```suggestion
            size={IconSize.Md}
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+    variant={TextVariant.BodySm}
+    fontWeight={FontWeight.Medium}
+    color={TextColor.TextAlternative}
+  >
+    Today
+  </Text>
+);
+
+/** Pill badge for `titleEndAccessory` demos (not in reference screenshot). */
+const TestnetBadge = () => (
```
amandaye0h: Imo this testnet badge shouldn't be part of the docs, since it's visually dissimilar from the rest of the components in our library. We can include it temporarily, but it should be updated to a component with squareish borders.

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+      title={USDC_TITLE}
+      subtitle={USDC_SUBTITLE}
+      amount={USDC_AMOUNT}
+      bottomAccessory={
+        <Box
+          flexDirection={BoxFlexDirection.Row}
+          alignItems={BoxAlignItems.Center}
+          gap={1}
+        >
+          <Icon name={IconName.SecurityAlert} size={IconSize.Xs} />
```
amandaye0h:
```suggestion
          <Icon name={IconName.SecurityAlert} size={IconSize.Sm} />
```
Xs is too small

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+  </Text>
+);
+
+/** Pill badge for `titleEndAccessory` demos (not in reference screenshot). */
+const TestnetBadge = () => (
+  <Box
+    flexDirection={BoxFlexDirection.Row}
+    alignItems={BoxAlignItems.Center}
+    gap={1}
+    twClassName="rounded-full bg-warning-muted px-2 py-1"
```
amandaye0h, line 53: This should be replaced with Tag in the future

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,171 @@
+// Third party dependencies.
+import { isReactNodeRenderable } from '@metamask/design-system-shared';
```
georgewrmarshall: Blocker: `isReactNodeRenderable` is being removed from `design-system-shared` in #1076. This import will cause a build failure once that PR merges.

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,171 @@
...
+    isReactNodeRenderable(subtitle) ||
+    isReactNodeRenderable(subtitleEndAccessory);
+  const renderAmountRow =
+    isReactNodeRenderable(amount) || isReactNodeRenderable(amountEndAccessory);
+  const renderBottomLabelRow =
+    isReactNodeRenderable(bottomLabel) ||
+    isReactNodeRenderable(bottomLabelEndAccessory);
+  const renderBottomAccessory =
+    !renderBottomLabelRow && isReactNodeRenderable(bottomAccessory);
+
```
georgewrmarshall: This entire block — the four `*Node` normalization variables and the five `render*` booleans — should be removed.

The normalization variables coerce `false`/`null` to `undefined` before passing to `BoxRow.endAccessory`. `BoxRow` receives a `ReactNode` prop; React renders all three values as nothing. Pass the props directly.

The `render*` booleans are doing what React already does natively — `{null}`, `{undefined}`, and `{false}` in JSX produce no output.

Additonally, the `||` pattern allows an accessory to render without its primary content. `renderTitleRow` is true when `titleEndAccessory` alone is truthy, producing a `BoxRow` with a floating accessory and no title. An accessory is compositionally dependent on its primary — it has no standalone meaning. The guard should be the primary prop only:

```tsx
{title && (
  <BoxRow
    endAccessory={titleEndAccessory}
    textProps={{ variant: TextVariant.HeadingSm, color: TextColor.TextDefault, ...titleProps }}
  >
    {title}
  </BoxRow>
)}
{subtitle && (
  <BoxRow
    endAccessory={subtitleEndAccessory}
    textProps={{ variant: TextVariant.BodySm, fontWeight: FontWeight.Medium, color: TextColor.TextAlternative, ...subtitleProps }}
  >
    {subtitle}
  </BoxRow>
)}
```

For the bottom row, use an inline ternary rather than a pre-computed variable:

```tsx
{bottomLabel ? (
  <BoxRow
    endAccessory={bottomLabelEndAccessory}
    textProps={{ variant: TextVariant.BodySm, fontWeight: FontWeight.Medium, color: TextColor.TextAlternative, ...bottomLabelProps }}
  >
    {bottomLabel}
  </BoxRow>
) : bottomAccessory}
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,171 @@
...
+        fontWeight: FontWeight.Medium,
+        color: TextColor.TextAlternative,
+        ...subtitleProps,
+      }}
+    >
+      {subtitle}
+    </BoxRow>
+  );
+
+  const titleAvatarSlot = (
```
georgewrmarshall: `titleAvatarSlot`, `identityRow`, `amountRow`, and `bottomLabelRow` are each used exactly once with no conditional branching on whether to build them. Extracting single-use JSX into named variables adds indirection without adding clarity — it requires jumping between definition and usage to follow the layout, and makes bugs harder to spot (see the `align-middle` comment below and the redundant `render*` conditionals above). Inline these directly in the return so the layout is readable top-to-bottom.

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,171 @@
...
+      justifyContent={BoxJustifyContent.Center}
+    >
+      {titleAvatar}
+    </Box>
+  );
+
+  const identityRow = (
+    <BoxRow
+      gap={4}
+      twClassName="h-14 align-middle"
```
georgewrmarshall: We should check `align-middle` is a valid twrnc classname as per the cursor bugbot comment

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.types.ts
```diff
@@ -0,0 +1,33 @@
...
+// Internal dependencies.
+import type { TextProps } from '../Text/Text.types';
+
+/**
+ * TitleSubpage component props (React Native).
+ * Extends {@link TitleSubpagePropsShared} (requires `title` and `titleAvatar`) with platform `Text` passthroughs, `twClassName`, and `View` props.
+ */
+export type TitleSubpageProps = TitleSubpagePropsShared & {
+  /**
+   * Optional props merged into {@link BoxHorizontal} `textProps` when `amount` is a string.
```
georgewrmarshall: `{@link BoxHorizontal}` is stale — `BoxHorizontal` was renamed to `BoxRow` in 0.16.0. This appears in all four `*Props` JSDoc comments in this file. Update all four to `{@link BoxRow}`.

packages/design-system-react-native/src/components/TitleSubpage/README.md
```diff
@@ -0,0 +1,444 @@
...
+| TYPE        | REQUIRED | DEFAULT |
+| ----------- | -------- | ------- |
+| `ReactNode` | Yes      | —       |
+
+```tsx
+import { TitleSubpage, Box } from '@metamask/design-system-react-native';
+
+<TitleSubpage
+  titleAvatar={
+    <Box twClassName="h-8 w-8 shrink-0 rounded-full bg-primary-muted" />
```
georgewrmarshall: The `titleAvatar` example uses a plain `Box` placeholder. The prop description explicitly says to pass an avatar or token component such as `AvatarToken`. All `titleAvatar` usages throughout the README should use a real `AvatarToken` example rather than a placeholder `Box`.

packages/design-system-react-native/src/components/TitleSubpage/README.md
```diff
@@ -0,0 +1,444 @@
...
+  }
+  title="Send"
+  amount="$4.42"
+/>;
+```
+
+### `title`
+
+Title row (required). When `title` is a string, it uses `TextVariant.HeadingSm` and `TextColor.TextDefault` (merged with `titleProps`). Pass a `ReactNode` for custom layout. The row also renders when only `titleEndAccessory` is renderable (for example `title={false}` with an end accessory).
```
georgewrmarshall: "The row also renders when only `titleEndAccessory` is renderable (for example `title={false}` with an end accessory)" documents the `||` conditional behaviour that should be removed — see the implementation comment. An accessory should not render without its primary content. Remove this sentence. The same applies to the equivalent sentence in the `subtitle` section.

georgewrmarshall (changes requested):
Two architectural patterns run throughout the implementation that need to be addressed before merge.

**1. Conditional rendering logic React already handles**

The component pre-computes `renderTitleRow`, `renderSubtitleRow`, `renderAmountRow`, `renderBottomLabelRow`, and `renderBottomAccessory` before the return, and separately normalises all four `*EndAccessory` props to `undefined` when `isReactNodeRenderable` returns false. None of this is necessary — `null`, `undefined`, and `false` are valid `ReactNode` values that React renders as nothing. This is exactly the pattern we identified and removed from `HeaderRoot` in #1076 (where `isReactNodeRenderable` is also being deleted from `design-system-shared`, making this a build-time blocker once that lands).

**2. Indirection that makes the layout harder to read and easier to miss bugs in**

`titleAvatarSlot`, `titleRow`, `subtitleRow`, `identityRow`, `amountRow`, and `bottomLabelRow` are all extracted into variables and then referenced exactly once with no conditional branching. JSX is readable top-to-bottom — pulling single-use fragments into named variables forces readers to jump between definition sites and the return, obscuring the actual layout and making redundant conditionals harder to spot.

Left some other more minor comments

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,132 @@
+// Third party dependencies.
+import React from 'react';
+
+// Internal dependencies.
+import { BoxAlignItems, BoxJustifyContent } from '../../types';
+import { Box } from '../Box';
+import { BoxRow } from '../BoxRow';
+import { TextVariant, TextColor, FontWeight } from '../Text';
```
georgewrmarshall: suggestion: `TextVariant`, `TextColor`, and `FontWeight` should be imported from `@metamask/design-system-shared`, not from `../Text`. These are shared types owned by the shared package — importing through a sibling component creates false semantic coupling (see `component-architecture.md` → "Where to Import Shared Types").

```suggestion
import { TextVariant, TextColor, FontWeight } from '@metamask/design-system-shared';
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,132 @@
...
+  bottomLabelEndAccessory,
+  amountProps,
+  titleProps,
+  subtitleProps,
+  bottomLabelProps,
+  twClassName = '',
+  ...props
+}) => {
+  return (
+    <Box twClassName={twClassName} {...props}>
```
georgewrmarshall: suggestion: The root `Box` has no `gap` prop, so the identity row, amount row, and bottom rows render with zero vertical spacing — they'll stack flush. For reference, `TitleStandard` (PR #1051) uses `gap={1}` on its root `Box` to space its rows. Is the missing gap intentional here, or should it match?

```suggestion
    <Box twClassName={twClassName} gap={1} {...props}>
```

packages/design-system-shared/src/types/TitleSubpage/TitleSubpage.types.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * TitleSubpage component shared props (ADR-0004).
+ * Platform-independent properties; platform packages extend with `ViewProps` / `className`,
+ * `twClassName`, and platform `Text` prop passthroughs.
+ */
+export type TitleSubpagePropsShared = {
+  /**
+   * Optional primary amount line below the title and optional subtitle (for example a fiat or token value).
+   * When a string, platforms typically wrap with large display styles via `textProps`.
+   * The amount row renders when `amount` or `amountEndAccessory` is renderable.
```
georgewrmarshall: suggestion: The JSDoc says "The amount row renders when `amount` or `amountEndAccessory` is renderable" but the implementation (correctly) only checks `amount`. Same for `bottomLabel` / `bottomLabelEndAccessory` on lines 42 and 48. The JSDoc should be updated to match the implementation so consumers aren't misled.

packages/design-system-react-native/src/components/TitleSubpage/README.md
```diff
@@ -0,0 +1,469 @@
...
+
+```tsx
+import {
+  AvatarBaseSize,
+  AvatarToken,
+  TitleSubpage,
+} from '@metamask/design-system-react-native';
+
+<TitleSubpage
+  titleAvatar={<AvatarToken name="ethereum" size={AvatarBaseSize.Lg} />}
```
georgewrmarshall: suggestion: `AvatarBaseSize` is the generic base type — `AvatarToken` has its own `AvatarTokenSize` export (which the stories already use correctly). All README examples use `AvatarBaseSize` instead of `AvatarTokenSize`. This appears in every `titleAvatar` example throughout the file (~30 instances). Should be `AvatarTokenSize` everywhere to match the component's own API.

```suggestion
  titleAvatar={<AvatarToken name="ethereum" size={AvatarTokenSize.Lg} />}
```

Also update the import on line 7 (and all equivalent imports throughout) from `AvatarBaseSize` to `AvatarTokenSize`.

packages/design-system-react-native/src/components/TitleSubpage/README.md
```diff
@@ -0,0 +1,469 @@
...
+  title="Send"
+  titleEndAccessory={
+    <Box twClassName="ml-2">
+      <Icon name={IconName.Info} size={IconSize.Sm} />
+    </Box>
```
georgewrmarshall: suggestion: `Icon` accepts `twClassName` directly — the wrapping `Box` is unnecessary. Also, `BoxRow` already applies `gap={1}` between its children and the `endAccessory`. Is the extra `ml-2` even needed, or should the `BoxRow` gap be adjusted instead?

This same pattern appears in 3 places in the README: `titleEndAccessory` (here), `subtitleEndAccessory` (line 145), and `amountEndAccessory` (line 199).

```suggestion
    <Icon name={IconName.Info} size={IconSize.Sm} twClassName="ml-2" />
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
```diff
@@ -0,0 +1,235 @@
...
+      control: 'text',
+    },
+    bottomLabel: {
+      control: 'text',
+    },
+    twClassName: { control: 'text' },
+  },
+  decorators: [
+    (Story) => (
+      <Box twClassName="w-full bg-background-default p-4">
```
georgewrmarshall: non-blocking: Does this `Box` decorator reflect how this component will be placed in the Page/Screen in mobile? I think in general we should display the component without any padding/background color changes but if this is more reflective of how it will work in mobile I think that's ok.

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
```diff
@@ -0,0 +1,235 @@
...
+      subtitle={USDC_SUBTITLE}
+      amount={USDC_AMOUNT}
+      amountEndAccessory={
+        <Box twClassName="ml-1">
+          <Icon
+            name={IconName.Info}
+            size={IconSize.Md}
+            color={IconColor.IconAlternative}
+          />
+        </Box>
```
georgewrmarshall: suggestion: `Icon` accepts `twClassName` directly — no need for the `Box` wrapper. Same applies to `subtitleEndAccessory` (line 174).

```suggestion
      amountEndAccessory={
        <Icon
          name={IconName.Info}
          size={IconSize.Md}
          color={IconColor.IconAlternative}
          twClassName="ml-1"
        />
      }
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
```diff
@@ -0,0 +1,235 @@
+import type { Meta, StoryObj } from '@storybook/react-native';
+import React from 'react';
+
+import UsdcSVG from '../../assets/token-icons/usdc.svg';
+import { AvatarToken, AvatarTokenSize } from '../AvatarToken';
+import { Box, BoxAlignItems, BoxFlexDirection } from '../Box';
+import { Icon, IconName, IconSize, IconColor } from '../Icon';
+import { Text, TextColor, FontWeight, TextVariant } from '../Text';
```
georgewrmarshall: suggestion: Same as the component — `TextColor`, `FontWeight`, and `TextVariant` should be imported from `@metamask/design-system-shared`, not from `../Text`.

```suggestion
import { TextColor, FontWeight, TextVariant } from '@metamask/design-system-shared';
```

packages/design-system-react-native/src/components/TitleSubpage/README.md
```diff
@@ -0,0 +1,469 @@
...
+/>;
+```
+
+### `amountProps`
+
+Optional props merged into the amount `Text` when `amount` is a string. Use for `testID` or typography overrides.
+
+| TYPE                 | REQUIRED | DEFAULT     |
+| -------------------- | -------- | ----------- |
+| `Partial<TextProps>` | No       | `undefined` |
```
georgewrmarshall: suggestion: The type table shows `Partial<TextProps>` but the actual type is `Omit<Partial<TextProps>, 'children'>`. Same for `titleProps` (line 336), `subtitleProps` (line 359), and `bottomLabelProps` (line 383).

georgewrmarshall (commented):
Good progress on the simplification — the inlined layout and removal of `isReactNodeRenderable` / `render*` booleans / `align-middle` all look great. A few remaining items, mostly the same patterns flagged on TitleStandard (#1051) and TitleHub (#1052).

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
georgewrmarshall: suggestion: `TextVariant`, `TextColor`, and `FontWeight` should be imported from `@metamask/design-system-shared`, not from `../Text`. These are shared types owned by the shared package — importing through a sibling component creates false semantic coupling (see `component-architecture.md` → "Where to Import Shared Types").

```suggestion
import { TextVariant, TextColor, FontWeight } from '@metamask/design-system-shared';
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
georgewrmarshall: suggestion: The root `Box` has no `gap` prop, so the identity row, amount row, and bottom rows render with zero vertical spacing — they'll stack flush. For reference, `TitleStandard` (PR #1051) uses `gap={1}` on its root `Box` to space its rows. Is the missing gap intentional here, or should it match?

```suggestion
    <Box twClassName={twClassName} gap={1} {...props}>
```

georgewrmarshall (commented):
Follow-up review — component structure, tests, and the same recurring patterns from TitleStandard/TitleHub.

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,132 @@
...
+        twClassName="h-14"
+        startAccessory={
+          // Title Avatar Slot
+          <Box
+            twClassName="h-10 w-10 shrink-0"
+            alignItems={BoxAlignItems.Center}
+            justifyContent={BoxJustifyContent.Center}
+          >
+            {titleAvatar}
+          </Box>
```
georgewrmarshall: suggestion: Is this `Box` wrapper necessary? `AvatarBase` (which `AvatarToken` extends) already handles its own sizing via `TWCLASSMAP_AVATARBASE_SIZE_DIMENSION` and centering via `items-center justify-center`. At `AvatarTokenSize.Lg` the avatar is already 40×40, so this wrapper duplicates that sizing. If the intent is to enforce a fixed 40×40 slot regardless of what avatar size the consumer passes, that's worth documenting — otherwise consider removing the wrapper and letting the avatar render directly.

```suggestion
            {titleAvatar}
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,132 @@
...
+                fontWeight: FontWeight.Medium,
+                color: TextColor.TextAlternative,
+                ...subtitleProps,
+              }}
+            >
+              {subtitle}
+            </BoxRow>
+          )}
+        </Box>
+      </BoxRow>
```
georgewrmarshall, line 101: non-blocking: The identity row `BoxRow` (line 56), the title/subtitle column `Box` (line 71), and the bottom label row `BoxRow` (line 114) are not accessible via prop objects. This breaks the principle of allowing consumer access to all child layout components — the pattern used by `titleWrapperProps` / `amountWrapperProps` in TitleHub. Unlikely to be needed now, but worth noting for consistency. Consider exposing `identityRowProps`, `titleColumnProps`, and `bottomLabelWrapperProps` (the latter already exists on TitleHub/TitleStandard).

packages/design-system-shared/src/types/TitleSubpage/TitleSubpage.types.ts
georgewrmarshall: suggestion: The JSDoc says "The amount row renders when `amount` or `amountEndAccessory` is renderable" but the implementation (correctly) only checks `amount`. Same for `bottomLabel` / `bottomLabelEndAccessory` on lines 42 and 48. The JSDoc should be updated to match the implementation so consumers aren't misled.

packages/design-system-react-native/src/components/TitleSubpage/README.md
georgewrmarshall: suggestion: `AvatarBaseSize` is the generic base type — `AvatarToken` has its own `AvatarTokenSize` export (which the stories already use correctly). All README examples use `AvatarBaseSize` instead of `AvatarTokenSize`. This appears in every `titleAvatar` example throughout the file (~30 instances). Should be `AvatarTokenSize` everywhere to match the component's own API.

```suggestion
  titleAvatar={<AvatarToken name="ethereum" size={AvatarTokenSize.Lg} />}
```

Also update the import on line 7 (and all equivalent imports throughout) from `AvatarBaseSize` to `AvatarTokenSize`.

packages/design-system-react-native/src/components/TitleSubpage/README.md
georgewrmarshall: suggestion: `Icon` accepts `twClassName` directly — the wrapping `Box` is unnecessary. Also, `BoxRow` already applies `gap={1}` between its children and the `endAccessory`. Is the extra `ml-2` even needed, or should the `BoxRow` gap be adjusted instead?

This same pattern appears in 3 places in the README: `titleEndAccessory` (here), `subtitleEndAccessory` (line 145), and `amountEndAccessory` (line 199).

```suggestion
    <Icon name={IconName.Info} size={IconSize.Sm} twClassName="ml-2" />
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
georgewrmarshall: non-blocking: Does this `Box` decorator reflect how this component will be placed in the Page/Screen in mobile? I think in general we should display the component without any padding/background color changes but if this is more reflective of how it will work in mobile I think that's ok.

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
georgewrmarshall: suggestion: `Icon` accepts `twClassName` directly — no need for the `Box` wrapper. Same applies to `subtitleEndAccessory` story (line 174).

```suggestion
      amountEndAccessory={
        <Icon
          name={IconName.Info}
          size={IconSize.Md}
          color={IconColor.IconAlternative}
          twClassName="ml-1"
        />
      }
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
georgewrmarshall: suggestion: Same as the component — `TextColor`, `FontWeight`, and `TextVariant` should be imported from `@metamask/design-system-shared`, not from `../Text`.

```suggestion
import { TextColor, FontWeight, TextVariant } from '@metamask/design-system-shared';
```

packages/design-system-react-native/src/components/TitleSubpage/README.md
georgewrmarshall: suggestion: The type table shows `Partial<TextProps>` but the actual type is `Omit<Partial<TextProps>, 'children'>`. Same for `titleProps` (line 336), `subtitleProps` (line 359), and `bottomLabelProps` (line 383).

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.test.tsx
```diff
@@ -0,0 +1,497 @@
...
+          titleAvatar={defaultTitleAvatar}
+          title="Hi"
+          amount="$4.42"
+          titleEndAccessory={false}
+        />,
+      );
+
+      expect(getByText('Hi')).toBeOnTheScreen();
+      expect(getByText('$4.42')).toBeOnTheScreen();
+    });
```
georgewrmarshall, line 231: suggestion: This test is named "renders title only" implying `titleEndAccessory` is excluded when `false`, but it only asserts that title and amount are present — it never verifies the accessory is absent. The assertion doesn't match the test name.

Same issue with "when amountEndAccessory is false → renders amount only" (lines 433-446) — asserts amount is present but never checks the accessory is absent.

Both tests should add a negative assertion, e.g.:

```tsx
// Add to "when titleEndAccessory is false"
const { getByText, queryByText } = render(...);
expect(queryByText('some accessory text')).not.toBeOnTheScreen();
```

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.stories.tsx
brianacnguyen: yes

georgewrmarshall (dismissed):
Looks strong after the latest round of changes.

A lot of the earlier structural feedback has been addressed well here:
- the component layout is much easier to follow now that the extra render indirection was removed
- the root spacing is in place and the overall structure reads clearly top-to-bottom
- the stories are cleaner and feel more representative after the simplification pass
- the README examples are in much better shape and more aligned with component conventions
- the tests cover the important conditional rendering and prop passthrough paths well
- overall this feels consistent with the direction we landed on for the other title components

A couple non-blocking follow-ups:

packages/design-system-react-native/src/components/TitleSubpage/README.md
```diff
@@ -0,0 +1,478 @@
+# TitleSubpage
+
+TitleSubpage lays out a required identity block (leading `titleAvatar` beside a title stack), an optional subtitle, an optional amount row, optional bottom rows, and optional inline accessories per row. On React Native, `titleAvatar` is passed as the `startAccessory` of the identity `BoxRow`; use `AvatarToken` at `AvatarTokenSize.Lg` for the design-spec 40×40 footprint.
```
georgewrmarshall: non-blocking: the README currently describes `titleAvatar` as matching the design-spec 40×40 footprint, but the implementation now just passes it through as the identity row `startAccessory`. That may be totally fine if sizing is expected to come from the avatar component itself, but the docs may want to be a little more precise about that contract.

packages/design-system-react-native/src/components/TitleSubpage/index.ts
```diff
@@ -0,0 +1,3 @@
+export type { TitleSubpagePropsShared } from '@metamask/design-system-shared';
```
georgewrmarshall: non-blocking: I am not sure we need to re-export `TitleSubpagePropsShared` from the DSRN component entrypoint. The shared prop types are mainly for internal cross-platform alignment, while package consumers should use `TitleSubpageProps`. This seems fine to clean up in a broader bulk pass.

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,144 @@
...
+        startAccessory={
+          <Box
+            testID="title-subpage-title-avatar-slot"
+            alignItems={BoxAlignItems.Center}
+            justifyContent={BoxJustifyContent.Center}
+            twClassName="size-10 overflow-hidden"
+          >
+            {titleAvatar}
+          </Box>
+        }
```
georgewrmarshall: Still curious if this Box wrapper is necessary?

brianacnguyen: @georgewrmarshall this is to make sure the avatar is contained within a 40px x 40px box. what do you think?

georgewrmarshall: I think the component would be more flexible without the wrapper? We should ensure `AvatarToken` is size large but what if we need to add a badge or change something that goes outside of the `Box` it will be cut off or confined to the 40px size?

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
```diff
@@ -0,0 +1,145 @@
...
```
georgewrmarshall: suggestion: I think `titleAvatar` is more flexible if we pass it straight through as the `startAccessory` instead of wrapping it in a fixed `size-10 overflow-hidden` `Box`.

```tsx
startAccessory={titleAvatar}
```

We should still document `AvatarTokenSize.Lg` as the standard case, but enforcing that with a wrapper makes the slot less composable. If the caller needs to layer something on top of the avatar, like a network badge, the wrapper clips it or confines it to the 40px box.

The storybook example below shows the kind of composition this blocks:

```tsx
const StoryTitleAvatar = () => (
  <BadgeWrapper
    badge={
      <AvatarNetwork src={EthSvg} size={AvatarTokenSize.Xs} name="USD Coin" />
    }
  >
    <AvatarToken src={UsdcSVG} size={AvatarTokenSize.Lg} name="USD Coin" />
  </BadgeWrapper>
);
```

### Before

Screenshot 2026-04-23 at 12 14 11 PM

### After

Screenshot 2026-04-23 at 12 17 44 PM


I think the component should stay agnostic here and let the caller own the avatar composition. If we want to preserve the design guidance, we can keep `AvatarTokenSize.Lg` in the docs and story examples rather than constraining the rendered node in the component itself.

georgewrmarshall (commented):
The component looks in a good place after the review rounds. I just have one remaining suggestion around keeping the `titleAvatar` slot more flexible for composed avatar use cases.

packages/design-system-shared/src/types/TitleSubpage/TitleSubpage.types.ts
```diff
@@ -0,0 +1,55 @@
...
+   * the `startAccessory` of the identity `BoxRow` (typically size the avatar to the design-spec 40×40 slot).
+   */
+  titleAvatar: ReactNode;
+  /**
+   * Optional accessory rendered inline to the right of `title` in the title row.
+   */
+  titleEndAccessory?: ReactNode;
+  /**
+   * Optional subtitle row below the title and above the amount (via platform `textProps` when a string).
+   * The subtitle row renders when `subtitle` is renderable.
```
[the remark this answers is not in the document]
georgewrmarshall: false positive we always provide content in a string

packages/design-system-react-native/src/components/TitleSubpage/TitleSubpage.tsx
brianacnguyen: Convinced. Let me update

packages/design-system-react-native/src/components/TitleSubpage/README.md
```diff
@@ -0,0 +1,512 @@
+# TitleSubpage
+
+TitleSubpage lays out a required identity block (leading `titleAvatar` beside a title stack), an optional subtitle, an optional amount row, optional bottom rows, and optional inline accessories per row. On React Native, `titleAvatar` is passed straight through as the identity `BoxRow` `startAccessory` (no fixed-size or `overflow: hidden` wrapper), so network badges and other overlays can extend past the token without being clipped. For the default layout, use `AvatarToken` at `AvatarTokenSize.Lg` (40×40) so the token aligns with the row; the component stays agnostic—you own composition when you need badges or other chrome.
```
georgewrmarshall, line 3: non-blocking: I’d simplify this opening description to keep it consumer-focused. Something like: `TitleSubpage` lays out a required identity block with a leading `titleAvatar` beside a title stack, plus an optional subtitle, optional amount row, optional bottom rows, and optional inline accessories per row. For avatars passed to `titleAvatar`, use a large size, for example `AvatarToken` at `AvatarTokenSize.Lg` (40×40).

georgewrmarshall (approved):
Approving. The component looks in a good place after the recent changes.

I just have one non-blocking docs follow-up. I’ll also adjust the Cursor documentation rule in an effort to keep docs consumer-focused and avoid process history, implementation-detail narration, or extra fluff in these README additions.

PR #1072, opened 2026-04-11
fix: resolve transitive content scanning requirement for design-system-shared in tailwind-preset

### LINKED ISSUE #1071: feat: safelist shared typography utility classes in tailwind-preset to remove transitive content scanning requirement
## Summary

From `@metamask/design-system-react@0.17.0`, the typography type constants (`TextVariant`, `FontWeight`, `FontStyle`, `FontFamily`, `TextColor`) were moved to `@metamask/design-system-shared`. This means the Tailwind utility classes those constants reference now live in the compiled output of `design-system-shared`, not just `design-system-react`.

As a result, consumers must add a second `content` glob to their Tailwind config to prevent the JIT engine from purging those classes:

```js
// tailwind.config.js — required from design-system-react@0.17.0+
content: [
  './node_modules/@metamask/design-system-react/**/*.{mjs,cjs}',
  './node_modules/@metamask/design-system-shared/**/*.{mjs,cjs}', // new
]
```

This creates two problems:

1. **Fragile by default** — `design-system-shared` is a transitive dependency. If it isn't hoisted to the top-level `node_modules` (possible in strict hoisting environments or future yarn/pnpm changes), Tailwind silently skips the glob and typography classes are purged from the build with no error.

2. **Leaks internal package structure to consumers** — consumers shouldn't need to know which sub-package owns which class strings. That's an internal implementation detail of the design system.

## Proposed solution

Add the affected typography utility classes to the `safelist` in `@metamask/design-system-tailwind-preset`. Safelisted classes are always emitted regardless of content scanning, so consumers get them automatically when they use the preset — no `content` entry changes required now or in the future.

The safelist would cover the classes referenced by `FontWeight`, `FontStyle`, and `FontFamily` const values (a small, bounded set). For example:

```js
// inside design-system-tailwind-preset
safelist: [
  'font-bold',
  'font-medium',
  'font-regular',
  'italic',
  'not-italic',
  'font-default',
  'font-accent',
  'font-hero',
]
```

Any future migration of class-bearing constants to `design-system-shared` (or any other sub-package) would then be transparent to consumers — they bump the preset version and nothing in their Tailwind config changes.

## Context

- Discovered while upgrading `metamask-extension` to `design-system-react@0.17.0` — MetaMask/metamask-extension#41662
- The extension currently works around this by (a) pinning `design-system-shared` as a direct dependency and (b) adding the second content glob, but this is a maintenance burden that should be solved at the source.

## Alternatives considered

**Option B — Export a side-effectful JS module from `design-system-shared` with dummy class references** — works but is fragile and non-obvious.

**Option C — Document the required content glob in the preset README** — reduces surprise but doesn't fix the hoisting fragility.

Option A (safelist in preset) is the right fix: it encapsulates the design system's internal structure and gives consumers a stable, zero-config experience.

georgewrmarshall:
## **Description**

From `@metamask/design-system-react@0.17.0`, typography type constants (`TextVariant`, `FontWeight`, `FontStyle`, `FontFamily`, `TextColor`) were moved to `@metamask/design-system-shared`. Because `design-system-shared` is a transitive dependency, Tailwind's JIT engine could not reliably scan its compiled output — consumers had to manually add `@metamask/design-system-shared` to their `content` glob, which is fragile in strict hoisting environments (yarn PnP, pnpm, etc.).

This PR fixes the issue at the right layer: the `design-system-tailwind-preset`.

**Solution:**

Add `@metamask/design-system-shared` as a direct dependency of the preset and include its `dist` output in the preset's `content` array, resolved via `require.resolve` from the preset's own dependency graph:

```ts
const sharedDistGlob = `${require
  .resolve('@metamask/design-system-shared/package.json')
  .replace(/\/package\.json$/u, '')}/dist/**/*.{mjs,cjs}`;

const tailwindConfig: Config = {
  content: [sharedDistGlob],
  // ...
};
```

**Why `require.resolve` instead of a relative path string:**
Since `design-system-shared` is a direct dep of the preset, `require.resolve` finds it from the preset's own dep graph — not the consumer's. This is hoisting-safe regardless of how the consumer's `node_modules` is structured.

**Why not a safelist:**
A safelist would need to grow manually as more token-identity constants move into `design-system-shared` (e.g. Box color constants). The `content` scanning approach scales automatically — new constants are picked up without any changes to this file.

Also documents the architectural decision in `.cursor/rules/` (token-identity vs semantic-value const objects, and the content scanning mechanism).

## **Related issues**

Fixes: #1071

Related: MetaMask/metamask-extension#41662 — once consumers upgrade to this preset version, the manual `@metamask/design-system-shared` content glob and direct dependency added in that PR can be removed.

## **Manual testing steps**

1. Build the preset: `yarn workspace @metamask/design-system-tailwind-preset run build`
2. Build Storybook: `yarn workspace @metamask/storybook-react run build-storybook`
3. Confirm `TextColor` utility classes (e.g. `text-default`, `text-primary-default`, `text-error-default`) are present in the generated CSS without `design-system-shared` in the app's own `content` glob

### **Before**

Consumers required:
```js
content: [
  './node_modules/@metamask/design-system-react/**/*.{mjs,cjs}',
  './node_modules/@metamask/design-system-shared/**/*.{mjs,cjs}', // manual, fragile
]
```

### **After**

Consumers only need:
```js
content: [
  './node_modules/@metamask/design-system-react/**/*.{mjs,cjs}',
  // design-system-shared covered automatically by the preset
]
```

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes Tailwind `content` scanning behavior and package dependencies, which can affect generated CSS and downstream builds, but the change is narrowly scoped to the preset and covered by a new test.
> 
> **Overview**
> Fixes Tailwind JIT class generation for token-identity constants coming from `@metamask/design-system-shared` by making it a direct dependency of `@metamask/design-system-tailwind-preset` and adding a hoisting-safe `content` glob that resolves the shared package’s `dist` via `require.resolve`.
> 
> Adds a regression test asserting the preset’s `content` includes the shared `dist` path, and documents the *token-identity vs semantic-value* constant pattern plus the new scanning mechanism in `.cursor/rules`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b7617e9b23b02c49d539020601b9203835117c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.cursor/rules/component-architecture.md
+++ b/.cursor/rules/component-architecture.md
@@ -114,10 +114,10 @@ Both React (Tailwind) and React Native (TWRNC) share the same design token class
 
 **Two patterns:**
 
-| Pattern | Example | Value is… | Used as… |
-| ------- | ------- | --------- | -------- |
-| Token identity | `TextColor.TextDefault = 'text-default'` | The class string itself (same on both platforms) | `className={color}` / `twClassName={color}` directly |
-| Semantic value | `FontWeight.Bold = 'bold'` | Abstract identifier (platforms need different class strings) | Mapped in platform `.constants.ts` |
+| Pattern        | Example                                  | Value is…                                                    | Used as…                                             |
+| -------------- | ---------------------------------------- | ------------------------------------------------------------ | ---------------------------------------------------- |
+| Token identity | `TextColor.TextDefault = 'text-default'` | The class string itself (same on both platforms)             | `className={color}` / `twClassName={color}` directly |
+| Semantic value | `FontWeight.Bold = 'bold'`               | Abstract identifier (platforms need different class strings) | Mapped in platform `.constants.ts`                   |
 
 ### Token identity constants (`TextColor`, `BoxBackgroundColor`, `BoxBorderColor`, etc.)
 
--- a/.cursor/rules/component-architecture.md
+++ b/.cursor/rules/component-architecture.md
@@ -106,6 +106,37 @@ Separate shared design system concerns from platform-specific implementation con
 - ✅ React: `className?: string` (platform layer)
 - ✅ React Native: `twClassName?: string` (platform layer)
 
+## Const Object Value Patterns in Shared
+
+Not all constants in `design-system-shared` have the same kind of value. The rule is simple: **if the class string is identical across both platforms, put it in shared; if the platforms need different class strings, use an abstract value in shared and map per platform.**
+
+Both React (Tailwind) and React Native (TWRNC) share the same design token class naming from their respective presets — so most token-identity constants can live in shared and be used directly as class names on both platforms.
+
+**Two patterns:**
+
+| Pattern | Example | Value is… | Used as… |
+| ------- | ------- | --------- | -------- |
+| Token identity | `TextColor.TextDefault = 'text-default'` | The class string itself (same on both platforms) | `className={color}` / `twClassName={color}` directly |
+| Semantic value | `FontWeight.Bold = 'bold'` | Abstract identifier (platforms need different class strings) | Mapped in platform `.constants.ts` |
+
+### Token identity constants (`TextColor`, `BoxBackgroundColor`, `BoxBorderColor`, etc.)
+
+- Values ARE the Tailwind/TWRNC class strings — identical across both platforms
+- Used directly: `className={color}` (React) / `twClassName={color}` (React Native)
+- ✅ Values live in shared and are used as-is on both platforms
+
+### Semantic constants (`FontWeight`, `FontStyle`, `FontFamily`, `TextVariant`)
+
+- Platforms need different class strings for these — they cannot share a single value:
+  - React: `FontWeight.Bold → 'font-bold'`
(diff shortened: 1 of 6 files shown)


georgewrmarshall:
Closing this PR — the approach is fundamentally incompatible with Tailwind CSS v3's preset behavior. Tailwind v3 does not merge `content` arrays from presets; the consumer's `content` completely replaces the preset's. The `sharedDistGlob` entry added here is silently discarded for all real consumers, making the fix a no-op. The storybook test passing was a false positive — `apps/storybook-react/tailwind.config.js` already has `design-system-shared` in its own `content` array directly.\n\nThe correct fix lives in the consumer (see MetaMask/metamask-extension#41662): consumers must add `@metamask/design-system-shared` to their own Tailwind `content` glob. This should be documented clearly in the preset's README and the migration guide for v32, not worked around at the preset layer.

PR #1076, opened 2026-04-13
refactor: simplify HeaderRoot slot guards to direct conditionals

### LINKED ISSUE #1075: refactor: remove isReactNodeRenderable utility — React handles empty values natively
## Description

`isReactNodeRenderable` was introduced in #1029 as a shared utility in `@metamask/design-system-shared`. It adds abstraction and package weight without solving a real problem — and introduced a subtle regression. A broader audit of `@metamask/design-system-react-native` found similar patterns across multiple components that are worth investigating. This issue tracks cleaning up all confirmed instances.

## Root cause

The original code used an explicit null check to decide whether to render children:

```tsx
children !== null && children !== undefined
```

This was unnecessary to begin with. React already handles every empty value uniformly — just rendering `{children}` is sufficient:

| Value | React renders |
|---|---|
| `null` | nothing |
| `undefined` | nothing |
| `false` | nothing |
| `true` | nothing |
| `''` | nothing visible |

You never need to guard or normalise a `ReactNode` before passing it to JSX. React handles it.

The only thing that needs a guard is **component logic** — deciding whether to render a container or row at all. For that, a plain `if (children)` is sufficient since JavaScript truthiness already covers `false`, `null`, and `undefined`.

## How the complexity snowballed

1. The original code used `children !== null && children !== undefined` — unnecessarily explicit, but harmless
2. Cursor Bugbot correctly flagged it didn't catch `false` from `{condition && <Comp />}` patterns
3. Rather than stepping back to `{children}` (which React handles correctly), the fix kept building on the wrong foundation
4. The reviewer suggested named guard variables for readability — the right idea
5. The author interpreted this as a shared exported utility — `isReactNodeRenderable` was created
6. The utility was then spread to `TitleHub`, and `|| null` normalisations were added throughout

The entire chain of complexity was a solution to a problem that never existed. `{children}` would have worked from day one.

## Why the utility is unnecessary

`!!prop` (and a plain `if (prop)`) already handles every realistic input correctly:

| Value | `!!prop` | `isReactNodeRenderable` |
|---|---|---|
| `false` (from `{cond && <Comp />}`) | `false` ✓ | `false` ✓ |
| `null` / `undefined` | `false` ✓ | `false` ✓ |
| `"text"` / `<Component />` | `true` ✓ | `true` ✓ |
| `0` (raw number) | `false` | `true` |
| `''` (empty string) | `false` | `true` ← caused a bug |

The utility only diverges from `!!` on `0` and `''`. Neither is a realistic input — consumers pass pre-formatted strings (`"$0.00"`, `"0 ETH"`), never raw numbers. The `''` divergence caused an immediate regression requiring a workaround:

```tsx
// band-aid needed because isReactNodeRenderable('') === true
const titleNode = isReactNodeRenderable(title) && title !== '' ? title : null;
```

## What the fix looks like

`renderLeftSection` in `HeaderRoot` went from:

```tsx
const hasRenderableChildren = isReactNodeRenderable(children);
const hasTitleContent = title !== false &&
  (isReactNodeRenderable(title) || isReactNodeRenderable(titleAccessory));
const shouldRenderTitleRow = !hasRenderableChildren && hasTitleContent;

const renderLeftSection = () => {
  if (hasRenderableChildren) return children;
  if (shouldRenderTitleRow) return <BoxRow ...>{isReactNodeRenderable(title) && title !== '' ? title : null}</BoxRow>;
  return null;
};
```

To:

```tsx
const renderLeftSection = () => {
  if (children) return children;
  if (title || titleAccessory) return <BoxRow ...>{title}</BoxRow>;
  return null;
};
```

## Audit — instances to investigate in `@metamask/design-system-react-native`

The following were identified as potentially applying the same unnecessary guard pattern. Each should be investigated before changing — some may have intentional reasons.

### Likely the same issue

- [ ] `TitleHub.tsx` — `isReactNodeRenderable(x)` throughout (tracked in #1076)
- [ ] `BannerBase.tsx:23` — `const hasContent = (content) => content !== null && content !== undefined`
- [ ] `Skeleton.tsx:21` — `const hasChildren = children !== null && children !== undefined`

### Investigate — may have intentional reasons

- [ ] `Toast.tsx:277` — `startAccessory !== null && startAccessory !== undefined && isValidElement(startAccessory)` — `isValidElement` check may be intentional
- [ ] `ListItem.tsx:62` — `Children.toArray(children).filter(Boolean)` — filtering may be intentional
- [ ] `BottomSheetFooter.tsx:26` — `Boolean(primaryButtonProps) && Boolean(secondaryButtonProps)`
- [ ] `HeaderBase.tsx:58` — `Boolean(hasAnyAccessory)` / `Boolean(hasStartContent)`
- [ ] `TextFieldSearch.tsx:60` — `Boolean(value) && clearButton`

## Remaining work — shared package cleanup

- [ ] Delete `packages/design-system-shared/src/utils/isReactNodeRenderable.ts`
- [ ] Delete `packages/design-system-shared/src/utils/isReactNodeRenderable.test.ts`
- [ ] Remove the export from `packages/design-system-shared/src/index.ts`
- [ ] Remove `react` + `@types/react` devDependencies from `packages/design-system-shared/package.json`

## References

- Introduced in: #1029
- Partial fix: #1076
- Current callsites: `TitleHub.tsx`, `BannerBase.tsx`, `Skeleton.tsx`, `Toast.tsx`, `BottomSheetFooter.tsx`, `HeaderBase.tsx`, `TextFieldSearch.tsx`, `ListItem.tsx`

georgewrmarshall:
## **Description**

`HeaderRoot` was using `isReactNodeRenderable` from `@metamask/design-system-shared` and several intermediate named boolean guards (`hasRenderableChildren`, `hasTitleContent`, `shouldRenderTitleRow`) to decide what to render in the left section. This logic was over-engineered — React already treats `null`, `undefined`, `false`, and `''` as nothing when rendering, so `!!prop` is the correct and sufficient check.

The guards also introduced a real bug: `isReactNodeRenderable('')` returns `true` (it only excludes `null`, `undefined`, and booleans), which forced an additional `&& title !== ''` workaround inline. Replacing everything with direct `if (children)` / `if (title)` conditionals removes the utility dependency, the workarounds, and the accidental complexity.

**Changes:**

- Removed `isReactNodeRenderable` import and all intermediate guard variables from `HeaderRoot`
- Replaced with direct `if (children)` / `if (title)` conditionals in `renderLeftSection`
- Removed the inner `Box` wrapper around the left section; `BoxRow` now renders directly as a flex child with `twClassName="flex-1"`
- End section container gains `ml-auto` to push it to the right without the wrapper
- Removed stale tests that asserted `titleAccessory` renders without a `title` (behaviour that was never intentional and is now correctly absent)
- Updated `MIGRATION.md` to document both breaking changes for 0.19.0

**Breaking changes:**

1. `titleAccessory` no longer renders when `title` is falsy — it is decoration on the title row, not a standalone slot
2. The intermediate `Box` wrapper around the left section is removed (structural change, no visual impact)

Related issue tracking the full `isReactNodeRenderable` cleanup: #1075

## **Related issues**

Fixes: #1075 (partial — `HeaderRoot` only; `TitleHub` and shared package cleanup tracked in the issue)

## **Manual testing steps**

1. Run `yarn storybook:ios` or `yarn storybook:android`
2. Navigate to `Components/HeaderRoot`
3. Verify all existing stories render correctly (Default, Title, TitleAccessory, Children, EndAccessory, EndButtonIconProps, EndButtonIconPropsMultiple)
4. Verify `titleAccessory` renders alongside `title` and does not render when `title` is omitted

### **Before**

No visual changes

https://github.com/user-attachments/assets/7e6e0db4-0a45-4a16-8b47-2a4016cd8f64

### **After**

https://github.com/user-attachments/assets/ea45dfb5-8a09-4b03-9504-c918347df7d5

---

> [!NOTE]
> **Medium Risk**
> Introduces breaking behavior changes in `HeaderRoot` (no standalone `titleAccessory`) and removes a previously exported shared utility, which can break downstream builds/usages if relied upon.
> 
> **Overview**
> **Simplifies `HeaderRoot` rendering logic and layout.** Replaces `isReactNodeRenderable`-based guards with direct `children`/`title` conditionals, removes the left-section wrapper `Box`, and uses flex utilities (including `ml-auto`) to keep the end section aligned.
> 
> **Behavior changes are now explicit and documented.** `titleAccessory` only renders when `title` is truthy (tests updated accordingly), Storybook examples were adjusted for truncation, and `MIGRATION.md` adds guidance for the 0.19.0 change.
> 
> **Shared package cleanup.** Removes the `isReactNodeRenderable` utility (and its tests) from `@metamask/design-system-shared` exports and adds a shared migration guide noting the removal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08a00393b274c7e01f0bda7075d2dbe4016cd69d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
```diff
@@ -90,11 +78,9 @@ export const HeaderRoot = ({
       testID={testID}
       {...viewProps}
     >
-      <Box alignItems={BoxAlignItems.Start} style={{ flex: 1 }}>
-        {renderLeftSection()}
-      </Box>
-      {hasEndSection ? (
-        <Box flexDirection={BoxFlexDirection.Row} gap={2}>
+      {renderLeftSection()}
```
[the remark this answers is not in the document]
georgewrmarshall: This is intentional the layout is still the same. `children` should not have a wrapper

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
```diff
@@ -54,15 +45,11 @@ export const HeaderRoot = ({
...
-
   const renderLeftSection = () => {
-    if (hasRenderableChildren) {
+    if (children) {
       return children;
     }
-    if (shouldRenderTitleRow) {
-      const titleNode =
-        isReactNodeRenderable(title) && title !== '' ? title : null;
+    if (title || titleAccessory) {
```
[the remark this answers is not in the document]
georgewrmarshall: Agreed. Will update this, it should only render if title exists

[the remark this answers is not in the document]
georgewrmarshall: Updated: https://github.com/MetaMask/metamask-design-system/pull/1076/commits/96b0c4ac9269d54a0159bb2f088f9bc888f7b830

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
```diff
@@ -54,24 +45,21 @@ export const HeaderRoot = ({
...
-  const hasEndSection = Boolean(endSectionContent);
-
   const renderLeftSection = () => {
-    if (hasRenderableChildren) {
+    if (children) {
       return children;
     }
-    if (shouldRenderTitleRow) {
-      const titleNode =
-        isReactNodeRenderable(title) && title !== '' ? title : null;
```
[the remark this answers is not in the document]
georgewrmarshall: This is intentional why would we render a `titleAccessory` without a `title`? If something like that is needed it can be rendered in it's own lock up in `children`

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
```diff
@@ -1,5 +1,4 @@
 // Third party dependencies.
-import { isReactNodeRenderable } from '@metamask/design-system-shared';
 import React from 'react';
```
georgewrmarshall, line 2: Removed `isReactNodeRenderable` from `@metamask/design-system-shared` — this was the only React Native component importing it. React already treats `null`, `undefined`, `false`, and `''` as nothing when rendering, so the utility added a shared package dependency and a type import without solving a real problem. The `!!prop` implicit coercion that plain `if (prop)` performs is correct and sufficient for all realistic slot prop inputs.

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
```diff
@@ -27,14 +26,6 @@ export const HeaderRoot = ({
 }: HeaderRootProps) => {
   const insets = useSafeAreaInsets();
```
georgewrmarshall, line 27: The three guard variables that lived here (`hasRenderableChildren`, `hasTitleContent`, `shouldRenderTitleRow`) were computing booleans that `if (children)` and `if (title)` now express directly. The most problematic was `hasTitleContent`, which combined `isReactNodeRenderable(title) || isReactNodeRenderable(titleAccessory)`

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
```diff
@@ -54,24 +45,21 @@ export const HeaderRoot = ({
...
-
   const renderLeftSection = () => {
-    if (hasRenderableChildren) {
+    if (children) {
       return children;
     }
-    if (shouldRenderTitleRow) {
-      const titleNode =
-        isReactNodeRenderable(title) && title !== '' ? title : null;
+    if (title) {
```
georgewrmarshall, line 52: `titleAccessory` is intentionally tied to `title` here — it is an accessory to the title row, not an independent slot. If you need a fully custom left section without a title string, use `children` instead. This makes the contract explicit: `children` owns its own layout, `title` + `titleAccessory` compose a managed row.

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
georgewrmarshall, line 81: The outer `Box` wrapper (with `style={{ flex: 1 }}`) has been removed — `children` is now responsible for its own layout, and `BoxRow` gets `twClassName="flex-1"` directly. Using inline `style` on a wrapper that children couldn't see or control was the wrong layer to express flex growth; moving it onto `BoxRow` keeps the responsibility with the node that actually needs it.

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.tsx
```diff
@@ -90,11 +78,9 @@ export const HeaderRoot = ({
...
       {...viewProps}
     >
-      <Box alignItems={BoxAlignItems.Start} style={{ flex: 1 }}>
-        {renderLeftSection()}
-      </Box>
-      {hasEndSection ? (
-        <Box flexDirection={BoxFlexDirection.Row} gap={2}>
+      {renderLeftSection()}
+      {endSectionContent ? (
+        <Box flexDirection={BoxFlexDirection.Row} gap={2} twClassName="ml-auto">
```
georgewrmarshall, line 83: `ml-auto` replaces the old approach of letting the left section's `flex: 1` push the end section to the right. Without the wrapper providing flex growth, the end section needs to position itself — `ml-auto` in a flex row achieves exactly that and is the idiomatic Tailwind pattern.

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.test.tsx
```diff
@@ -78,31 +78,6 @@ describe('HeaderRoot', () => {
       expect(getByTestId(LEFT_CHILDREN_TEST_ID)).toBeOnTheScreen();
     });
```
georgewrmarshall, line 79: These two tests were asserting behaviour that was never intentionally designed. Removing the tests is the correct response because the behaviour they described was wrong, not because the feature was cut.

packages/design-system-react-native/MIGRATION.md
```diff
@@ -27,6 +28,53 @@ This guide provides detailed instructions for migrating your project from one ve
 
 ## Version Updates
 
+### From version 0.18.0 to 0.19.0
+
+#### HeaderRoot: `titleAccessory` no longer renders without `title`
```
georgewrmarshall, line 33: The `titleAccessory` standalone rendering change is a genuine breaking change worth calling out clearly here. The `children` escape hatch is the right migration path — it gives consumers full control over the left section layout without depending on the component's internal title row structure.

packages/design-system-shared/src/utils/isReactNodeRenderable.ts
```diff
@@ -1,17 +0,0 @@
...
- * `{condition && <Component />}` and `{condition && 'Label'}` do not leave a boolean
- * or spurious layout when the condition is false. React does not render `true` or
- * `false` as visible output.
- *
- * @param node - A React child value (for example `children` or an optional slot prop).
- * @returns `true` when the value should be treated as renderable; otherwise `false`.
- */
-export function isReactNodeRenderable(node: ReactNode): boolean {
-  return node !== null && node !== undefined && typeof node !== 'boolean';
-}
```
georgewrmarshall, line 17: The utility file and its test are fully deleted rather than deprecated — there is no transition period because `!!prop` is a zero-effort replacement. Keeping the utility would maintain a public API commitment for logic that plain JavaScript already provides, and the shared package is the wrong place for React rendering semantics anyway.

packages/design-system-shared/src/utils/isReactNodeRenderable.test.ts
```diff
@@ -1,51 +0,0 @@
...
-  it('returns false for true', () => {
-    expect(isReactNodeRenderable(true)).toBe(false);
-  });
-
-  it('returns true for empty string', () => {
-    expect(isReactNodeRenderable('')).toBe(true);
-  });
-
-  it('returns true for zero', () => {
-    expect(isReactNodeRenderable(0)).toBe(true);
```
georgewrmarshall, line 27: The test suite is worth noting in the diff: it passed for `isReactNodeRenderable('')` returning `true` and `isReactNodeRenderable(0)` returning `true` — both treated as intended behaviour. In practice, consumers always pass pre-formatted strings (e.g. `'$0.00'`, not `0`) and empty string titles should never render a title row. The tests were accurately describing the utility's behaviour; the behaviour itself was the problem.

packages/design-system-shared/src/index.ts
```diff
@@ -6,8 +6,6 @@ export {
   generateIconSeed,
 } from './utils/caip-address';
 
-export { isReactNodeRenderable } from './utils/isReactNodeRenderable';
-
 // AvatarBase types (ADR-0003 + ADR-0004)
```
georgewrmarshall, line 9: Removing the export here is the only source-breaking change for consumers of `@metamask/design-system-shared`. Any downstream code that imports `isReactNodeRenderable` will get a compile-time or runtime resolution error, which is the intended signal to migrate. The MIGRATION.md documents the exact replacement.

packages/design-system-shared/MIGRATION.md
```diff
@@ -0,0 +1,47 @@
...
+
+### From version 0.11.0 to 0.12.0
+
+#### Removed: `isReactNodeRenderable` utility
+
+**What Changed:**
+
+`isReactNodeRenderable` has been removed from the public API. The utility was introduced to guard conditional slot props (`children`, `title`, `titleAccessory`) against falsy values, but it solves a problem React already handles natively — `null`, `undefined`, `false`, and `''` all render as nothing in JSX without any guard.
+
+The utility also introduced a subtle inconsistency: `isReactNodeRenderable('')` returned `true` (it only excluded `null`, `undefined`, and booleans), which required immediate workarounds like `&& title !== ''` at every callsite.
```
georgewrmarshall, line 20: First MIGRATION.md for `@metamask/design-system-shared`, following the same structure as the react-native and react packages (Table of Contents, Version Updates wrapper, What Changed / Migration / Impact sections). The empty string inconsistency is called out explicitly because it was the concrete evidence that the utility's design was wrong — not just unnecessary.

brianacnguyen (changes requested):
HeaderRoot should have title section flex-1 by design. Not sure your approach solves that

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.stories.tsx
```diff
@@ -83,15 +82,16 @@ export const Children = {
...
+          twClassName="flex-1" // flex-1 ensures the text is truncated
+        >
+          Imported Account 1 with a really long name
         </Text>
-        <Icon
-          name={IconName.ArrowDown}
-          color={IconColor.IconDefault}
-          twClassName="text-default"
-        />
+        <Icon name={IconName.ArrowDown} />
```
georgewrmarshall, line 94: `Text` and `Icon` default to `TextDefault` and `IconDefault` colors respectively — passing `twClassName="text-default"` and `color={IconColor.IconDefault}` was overriding defaults with themselves. Removed to keep the story clean and avoid implying these are necessary.

packages/design-system-react-native/src/components/HeaderRoot/HeaderRoot.stories.tsx
```diff
@@ -83,15 +82,16 @@ export const Children = {
...
       >
-        <Text variant={TextVariant.BodyMd} twClassName="text-default">
-          Imported Account 1
+        <Text
+          ellipsizeMode="tail"
+          numberOfLines={1}
+          twClassName="flex-1" // flex-1 ensures the text is truncated
+        >
+          Imported Account 1 with a really long name
         </Text>
```
georgewrmarshall, line 93: Updating custom component to show truncation

Image

brianacnguyen (approved):

PR #1081, opened 2026-04-15
refactor: Aligned TextField with cursor rules

brianacnguyen:
## **Description**

This pull request aligns **TextField** (`@metamask/design-system-react-native`) with our Cursor rules so the component and its docs/tests stay consistent with design-system conventions.

---

## **Related issues**

Fixes: N/A

---

## **Manual testing steps**

1. From the repo root, run **`yarn workspace @metamask/design-system-react-native run jest src/components/TextField/TextField.test.tsx --watchAll=false --coverage=false`** and confirm all tests pass.
2. Run **`yarn storybook:ios`** or **`yarn storybook:android`** (whichever you use for RN), open the **TextField** stories, and confirm default, error, disabled, accessories, and tap-to-focus behavior still match expectations.
3. *(Optional)* In a small local snippet, pass **`pressableProps={{ hitSlop: { top: 8, bottom: 8, left: 8, right: 8 } }}`** and confirm the larger touch target feels correct and does not break focus/blur.

---

### **After**
Everything is still functional

https://github.com/user-attachments/assets/aed72ccb-ec5b-48e2-91c7-27970306c223

---

---

---

> [!NOTE]
> **Medium Risk**
> This is a breaking public API refactor (prop layering, `ref` target change, removal of Pressable behavior) that can impact downstream call sites and imperative focus/testing selectors. Behavior changes to placeholder color precedence may also subtly affect UI theming expectations.
> 
> **Overview**
> **Refactors React Native `TextField`/`TextFieldSearch` to a layered API**: the root is now a `Box` (`View`) and native `TextInput` props must be passed via the new `inputProps` bag, with `ref` targeting the root container and a new `inputRef` for imperative access to the inner input. As part of this, the root is no longer a `Pressable` (tap-to-focus, `hitSlop`, and other Pressable-only props are removed), while focus/blur/error/disabled styling behavior is preserved.
> 
> Renames `isReadonly` to `isReadOnly` across `Input`, `TextField`, and `TextFieldSearch`, and changes `Input` so theme `placeholderTextColor` always overrides any caller-provided value. Updates Storybook stories, tests, READMEs, and `MIGRATION.md`, and introduces a shared `TextFieldPropsShared` type in `design-system-shared` to align cross-platform prop surfaces.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfac4d5e93a1dfd448e7b4220a30391ea68e2d40. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


georgewrmarshall (dismissed):
Nice update! One aspect of the component API that differs between React Native and React is that in React Native, `...props` is spread to the child input element, while the outer wrapper receives the `pressableProps` object. In Extension, it’s the other way around: [https://github.com/MetaMask/metamask-extension/blob/main/ui/components/component-library/text-field/text-field.tsx#L129-L158](https://github.com/MetaMask/metamask-extension/blob/main/ui/components/component-library/text-field/text-field.tsx#L129-L158). I think in general we follow the convention of spreading props to the outermost layer, so I’m not sure whether we want to adopt that here for alignment. It’s also how Material UI approaches their `TextField` component: [https://github.com/mui/material-ui/blob/master/packages/mui-material/src/TextField/TextField.js#L237](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/TextField/TextField.js#L237)

georgewrmarshall (commented):
A few non-blocking questions from comparing this refactor against the current RN `TextField` behavior and the extension `text-field` shape.

1. My understanding is that `Input` still exposes the native `TextInputProps` surface, including `placeholderTextColor`, but this change makes the theme placeholder color win over any caller-provided value. Is that intentional? If so, it may be worth documenting explicitly since direct `Input` consumers could reasonably expect native `placeholderTextColor` to still work.

2. My understanding is that this refactor removes the wrapper tap-to-focus behavior that the current RN `TextField` has, and that extension still has through the wrapper click handler. Is that intentional? I wanted to flag it because it changes the interaction model a bit and seems like a place where we may want stronger alignment across RN and extension if we can preserve it without compromising the native `TextInput` API.

3. On the docs side, I may be reading this wrong, but the README seems to suggest top-level accessibility props can be used on the root container, while the root is currently rendered with `accessible={false}`. Is the intent that those props live on the root container anyway, or should the docs steer people toward `inputProps` for the accessible input surface?

The `inputProps` direction still looks like the right long-term API to me, especially for things like `testID` and better alignment with extension. These comments are mainly about making sure the behavior changes here are intentional and clearly represented in the API/docs.

brianacnguyen:
> A few non-blocking questions from comparing this refactor against the current RN `TextField` behavior and the extension `text-field` shape.
> 
> 1. My understanding is that `Input` still exposes the native `TextInputProps` surface, including `placeholderTextColor`, but this change makes the theme placeholder color win over any caller-provided value. Is that intentional? If so, it may be worth documenting explicitly since direct `Input` consumers could reasonably expect native `placeholderTextColor` to still work.
> 2. My understanding is that this refactor removes the wrapper tap-to-focus behavior that the current RN `TextField` has, and that extension still has through the wrapper click handler. Is that intentional? I wanted to flag it because it changes the interaction model a bit and seems like a place where we may want stronger alignment across RN and extension if we can preserve it without compromising the native `TextInput` API.
> 3. On the docs side, I may be reading this wrong, but the README seems to suggest top-level accessibility props can be used on the root container, while the root is currently rendered with `accessible={false}`. Is the intent that those props live on the root container anyway, or should the docs steer people toward `inputProps` for the accessible input surface?
> 
> The `inputProps` direction still looks like the right long-term API to me, especially for things like `testID` and better alignment with extension. These comments are mainly about making sure the behavior changes here are intentional and clearly represented in the API/docs.

1/ Yes I don't think the placeholderTextColor should be overridable like it has previously. What do you think? Is there a legit case where we'd want to allow devs to override placeholderTextColor. The breaking change is documented in the migration doc I believe
2/ Overall users can still tap to focus with the TextField. The only difference is that, now, if they tap the start or end accessory, it won't be focused, which I think is fine. In most cases, the accessories will either be something static like an icon, which I think is fine to not have any press behavior their, or interactive, like button icons, which has its own actions anyway. The reason why I made this change is mainly to mirror the React counterpart (its currently wrapped with a Box) and I did check with AI and it doesnt seem necessary to have a wrapping Pressable. Additionally, I did look into the usage on mobile and no one is needing to use any props related to the Pressable, so the wrapper Pressable was there mainly for the focus interaction with the accessories, which I don't think is necessary, so replacing it with a wrapper Box is definitely an improvement
3/ IMO the docs should steer people toward the inputProps for the accessibile input surface. People using this component needs the input more than the box (hence the uplevel of some props)

packages/design-system-react-native/src/components/TextField/index.ts
```diff
@@ -1,2 +1,3 @@
 export { TextField } from './TextField';
-export type { TextFieldProps } from './TextField.types';
+export { type TextFieldPropsShared } from '@metamask/design-system-shared';
```
georgewrmarshall: Probably not necessary to export `TextFieldPropsShared`

packages/design-system-react-native/src/components/index.ts
```diff
@@ -196,7 +196,11 @@ export {
 export type { TextProps } from './Text';
 
 export { TextField } from './TextField';
-export type { TextFieldProps } from './TextField';
+export type {
+  TextFieldInputProps,
+  TextFieldProps,
+  TextFieldPropsShared,
```
georgewrmarshall: Not sure if we want to export `TextFieldPropsShared` as it's used internally and not meant for consumers?

georgewrmarshall (commented):
Looking good! Left one suggestion, we should also check with @amandaye0h. In the extension version of the `TextField` selecting the accessories initiate the focus state unless they're a button. You can test it in [the story](https://metamask.github.io/metamask-storybook/?path=/story/components-componentlibrary-textfield--start-accessory-end-accessory) I think we'd want the same behaviour for mobile

https://github.com/user-attachments/assets/5a61ed13-5220-4568-b7c1-b0c00bf09d3b

brianacnguyen:
> Looking good! Left one suggestion, we should also check with @amandaye0h. In the extension version of the `TextField` selecting the accessories initiate the focus state unless they're a button. You can test it in [the story](https://metamask.github.io/metamask-storybook/?path=/story/components-componentlibrary-textfield--start-accessory-end-accessory) I think we'd want the same behaviour for mobile
> 
>  focus.mov

If we want to align with this, it will make the component more complicated, especially since it will require either the wrapping container to use Pressable along with all the props associate it with. Is that worth it? IMO I don't think it's worth it since the extra Pressable is only there to mainly make sure the startAccessory can focus the input. Additionally, aside from search, there arent that many other inputs with startAccessory, so I don't think its worth the extra complexity IMO.

I did try the solution of keeping the Pressable while arranging the API around so we'd have some inputProps upleveled while keeping ...restProps for Pressable, and it just felt overly complicated for a case that's barely needed

georgewrmarshall (commented):
Ignore this review - posted prematurely

packages/design-system-react-native/src/components/TextField/index.ts
```diff
@@ -1,2 +1,2 @@
 export { TextField } from './TextField';
-export type { TextFieldProps } from './TextField.types';
+export type { TextFieldInputProps, TextFieldProps } from './TextField.types';
```
georgewrmarshall: suggestion: Do we want to export `TextFieldInputProps` as part of the public API? It seems mainly useful as the internal shape of `inputProps`, and exporting it expands the public surface for a fairly niche nested prop bag. Consumers can already derive the same shape from `TextFieldProps['inputProps']`, while a named export makes this lower-level split harder to change later without a breaking type change.

packages/design-system-react-native/src/components/TextField/TextField.types.ts
```diff
@@ -1,39 +1,76 @@
...
-   * Optional boolean to show the error state.
-   *
-   * @default false
+   * Called when the inner input loses focus (composed with TextField border state).
    */
-  isError?: boolean;
+  onBlur?: InputProps['onBlur'];
   /**
-   * Optional prop to replace the default Input with a custom element.
+   * Additional props for the inner `Input`. Use `accessibilityState={{ required: true }}` when
```
georgewrmarshall: suggestion: I think the required-field guidance needs another pass here. React Native's `TextInput` surface already gives us the native input props (https://reactnative.dev/docs/textinput), but `accessibilityState` there does not appear to include `required`, so `inputProps.accessibilityState={{ required: true }}` looks like it will point consumers toward something that does not typecheck. Since this PR is trying to clarify where input-level props belong, it would be good to switch this example to an accessibility pattern that matches the RN API.

packages/design-system-react-native/src/components/TextField/README.md
```diff
@@ -82,21 +194,50 @@ import { Text } from 'react-native';
...
+
+```tsx
+import { TextField } from '@metamask/design-system-react-native';
+
+<TextField value="" testID="my-text-field" placeholder="E2E" />;
+```
+
+### Layout and accessibility (`Box` / `View`)
+
+Pass `BoxProps` and React Native `View` props at the top level for layout and accessibility on the root container (for example `accessibilityHint`, `pointerEvents`). Keys reserved by TextField (`style`, `twClassName`, `testID`, `children`, `accessible`, and all keys owned by `TextFieldBaseProps`) are not passed through from this intersection. Prefer either Tailwind via `twClassName` or explicit `Box` layout props, and avoid conflicting layout when mixing both.
```
georgewrmarshall: suggestion: I think the accessibility docs are still a little inconsistent with the implementation. React Native describes `accessibilityHint` in the context of accessible elements (https://reactnative.dev/docs/accessibility), but `TextField` currently renders the root `Box` with `accessible={false}`, so the root does not appear to be the accessible surface in the current implementation. Since the intended API now seems to be 'outer container plus inner input', could the docs steer consumers toward `inputProps` for input accessibility unless we intend to support root accessibility here as well?

packages/design-system-shared/src/types/TextField/TextField.types.ts
```diff
@@ -0,0 +1,46 @@
...
+  value: string;
+  /**
+   * Called when the text changes. Uses React Native `TextInput` naming; web
+   * implementations may map this from the native input change event.
+   */
+  onChangeText?: (text: string) => void;
+  /** Placeholder shown when `value` is empty. */
+  placeholder?: string;
+  /** When true, the value cannot be edited. */
+  isReadonly?: boolean;
```
georgewrmarshall: non-blocking: Should this be `isReadOnly` rather than `isReadonly`? React Native's native prop is `readOnly` (https://reactnative.dev/docs/textinput), so if we want the design-system boolean form with an `is` prefix, `isReadOnly` feels more consistent with the platform spelling and a little easier to scan across platforms.

packages/design-system-react-native/src/components/TextField/TextField.tsx
```diff
@@ -4,87 +4,57 @@ import {
...
-      autoFocus = false,
+      startAccessory,
+      endAccessory,
+      style,
       twClassName,
-      onBlur,
-      onFocus,
       testID,
-      ...props
+      ...restProps
```
georgewrmarshall: non-blocking: Could we keep the spread name as `...props` instead of `...restProps` here? This component only peels off a relatively small owned set, and the older mobile/extension `TextField` implementations both use the simpler `props` naming. Keeping that here would make the component read a little more like the surrounding codebase and reduce naming noise in a public component.

packages/design-system-react-native/MIGRATION.md
```diff
@@ -91,6 +91,68 @@ These tokens had no backing CSS custom property, so any usage was already produc
...
+- **`placeholderTextColor`** is not supported on the public **`TextField`** API; the inner **`Input`** sets placeholder color from the theme.
+- Remaining top-level props on **`TextField`** are **`BoxProps`** (layout and **`View`** props from React Native), except for keys reserved by **`TextField`** (see type **`TextFieldProps`**). **`hitSlop`**, **`onPress`**, and other **`Pressable`**-only APIs are not supported on the root; tap-to-focus on the chrome is removed—users focus by tapping the **`Input`** / **`TextInput`**.
+- Cross-platform field definitions live in **`TextFieldPropsShared`** in **`@metamask/design-system-shared`**.
+- **`TextFieldSearchProps`** extends **`TextFieldProps`**; the same layering applies. **`onPressClearButton`**, **`clearButtonProps`**, **`startAccessory`**, **`endAccessory`**, and **`style`** behavior are unchanged.
+
+**Migration:**
+
+Move inner `TextInput` props from the root into **`inputProps`**. Keep **`placeholder`**, **`onFocus`**, and **`onBlur`** on the component root when you use them.
```
georgewrmarshall, line 112: suggestion: Since this API change is intentional, could we make the migration guidance more explicit about the new test ID contract? In the legacy mobile `TextField`, `testID` effectively targeted the inner input (https://github.com/MetaMask/metamask-mobile/blob/main/app/component-library/components/Form/TextField/TextField.tsx), while in this version `testID` targets the wrapper and `inputProps.testID` targets the actual `TextInput`. That split seems reasonable, but it is important enough for E2E and automation that it would be good to call out directly here and in the README.

packages/design-system-react-native/src/components/TextField/TextField.tsx
```diff
@@ -4,87 +4,57 @@ import {
...
     const [isFocused, setIsFocused] = useState(autoFocus);
     const inputRef = useRef<TextInput>(null);
     const tw = useTailwind();
 
+    const {
+      twClassName: inputTwClassNameFromProps,
+      ...inputRestWithoutTwClassName
+    } = inputProps ?? {};
+
     useImperativeHandle<TextInput | null, TextInput | null>(
```
georgewrmarshall: suggestion: Since `TextField` is a composite wrapper rather than a bare input, should we consider making the wrapper/input ref contract more explicit here by reserving the main ref for the wrapper and exposing a separate `inputRef` for the inner `TextInput`? React Native's `TextInput` does expose imperative methods like `.focus()` and `.blur()` (https://reactnative.dev/docs/textinput), so forwarding the main ref to the input is convenient today, but it also means we currently have no clean way to access the wrapper ref at all if we later need the container for measurement, layout, animation, or position/size reads. The extension `TextField` keeps that split explicit with `ref` + `inputRef` (https://github.com/MetaMask/metamask-extension/blob/main/ui/components/component-library/text-field/text-field.tsx), and that feels like a clearer contract for a composite component because it gives consumers access to both surfaces instead of spending the public ref on the internal input.

georgewrmarshall (commented):
I took another pass on this with the migration path in mind, especially current mobile and extension usage. The overall direction looks good, and a few of the earlier behavior questions now seem intentionally documented. The remaining feedback is mostly about tightening the public API shape and making the new contracts around accessibility, refs, and test targeting a little clearer before this settles into the shared surface.

georgewrmarshall:
> > Looking good! Left one suggestion, we should also check with @amandaye0h. In the extension version of the `TextField` selecting the accessories initiate the focus state unless they're a button. You can test it in [the story](https://metamask.github.io/metamask-storybook/?path=/story/components-componentlibrary-textfield--start-accessory-end-accessory) I think we'd want the same behaviour for mobile
> > focus.mov
> 
> If we want to align with this, it will make the component more complicated, especially since it will require either the wrapping container to use Pressable along with all the props associate it with. Is that worth it? IMO I don't think it's worth it since the extra Pressable is only there to mainly make sure the startAccessory can focus the input. Additionally, aside from search, there arent that many other inputs with startAccessory, so I don't think its worth the extra complexity IMO.
> 
> I did try the solution of keeping the Pressable while arranging the API around so we'd have some inputProps upleveled while keeping ...restProps for Pressable, and it just felt overly complicated for a case that's barely needed

Sounds good. Hopefully mobile is less sensitive to the inactive hit area of the TextInput. Let's monitor incase it becomes an issue

https://github.com/user-attachments/assets/97531e12-7a3f-48c0-b67f-421f45824093

georgewrmarshall (approved):

PR #1132, opened 2026-04-29
feat: `ModalFooter` migration (extension)

kirillzyusko:
## **Description**

Added `ModalFooter` migration to DSR.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-308

## **Manual testing steps**

1. Go to storybook
2. Check `ModalFooter` spec

### **Before**

image

### **After**

image

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are additive (new `ModalFooter` component, Storybook docs/stories, and tests) plus migration documentation updates, with no modifications to existing component behavior.
> 
> **Overview**
> Adds a new `ModalFooter` component to `@metamask/design-system-react`, rendering an optional primary/secondary action button row with a new `buttonsAlignment` option and enforcing button variants internally.
> 
> Exports `ModalFooter`/`ButtonsAlignment` from the package barrel, and includes full Storybook docs/stories plus unit tests. Updates `MIGRATION.md` with extension-to-design-system breaking-change mappings and before/after examples for `ModalFooter`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12ed5e4790c1b5f59ca2895a97792a40ad6faaaa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/design-system-react/src/components/ModalFooter/ModalFooter.tsx
+++ b/packages/design-system-react/src/components/ModalFooter/ModalFooter.tsx
@@ -1,6 +1,6 @@
+import { ButtonSize, ButtonVariant } from '@metamask/design-system-shared';
 import React, { forwardRef } from 'react';
 
-import { ButtonSize, ButtonVariant } from '../../types';
 import { twMerge } from '../../utils/tw-merge';
 import { Box, BoxAlignItems, BoxFlexWrap } from '../Box';
 import { Button } from '../Button';
--- a/packages/design-system-react/src/components/ModalFooter/ModalFooter.tsx
+++ b/packages/design-system-react/src/components/ModalFooter/ModalFooter.tsx
@@ -2,69 +2,65 @@ import { ButtonSize, ButtonVariant } from '@metamask/design-system-shared';
 import React, { forwardRef } from 'react';
 
 import { twMerge } from '../../utils/tw-merge';
-import { Box, BoxAlignItems, BoxFlexWrap } from '../Box';
+import { Box, BoxFlexDirection } from '../Box';
 import { Button } from '../Button';
 
+import { ButtonsAlignment } from './ModalFooter.types';
 import type { ModalFooterProps } from './ModalFooter.types';
-
-const FOOTER_BUTTON_CLASSNAME = 'flex-[1_0_auto]';
 
 export const ModalFooter = forwardRef<HTMLElement, ModalFooterProps>(
   (
     {
       className,
       children,
-      onSubmit,
-      submitButtonProps,
-      onCancel,
-      cancelButtonProps,
-      containerProps,
+      buttonsAlignment = ButtonsAlignment.Horizontal,
+      primaryButtonProps,
+      secondaryButtonProps,
       ...props
     },
     ref,
-  ) => (
-    <footer ref={ref} className={twMerge('px-4 pt-4', className)} {...props}>
-      {children}
-      <Box
-        flexWrap={BoxFlexWrap.Wrap}
-        alignItems={BoxAlignItems.Center}
-        gap={4}
-        {...containerProps}
-        className={twMerge(
-          'mx-auto flex max-w-[360px]',
-          containerProps?.className,
-        )}
-      >
-        {onCancel && (
-          <Button
-            size={ButtonSize.Lg}
-            variant={ButtonVariant.Secondary}
-            onClick={onCancel}
-            {...cancelButtonProps}
-            className={twMerge(
-              FOOTER_BUTTON_CLASSNAME,
-              cancelButtonProps?.className,
-            )}
+  ) => {
+    const isHorizontal = buttonsAlignment === ButtonsAlignment.Horizontal;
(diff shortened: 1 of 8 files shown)


packages/design-system-react/src/components/ModalFooter/ModalFooter.types.ts
```diff
@@ -0,0 +1,64 @@
...
+   * Optional custom content rendered above the built-in submit/cancel button row.
+   * Use this for inline form controls (e.g. a "do not show again" checkbox)
+   * or to fully replace the built-in button row by omitting `onSubmit` / `onCancel`.
+   */
+  children?: ReactNode;
+  /**
+   * Optional click handler for the submit button. When provided, a primary
+   * "submit" button is rendered.
+   */
+  onSubmit?: () => void;
```
georgewrmarshall: suggestion: Can we align this public API with React Native `BottomSheetFooter` before we ship a second footer contract? RN already uses `primaryButtonProps` / `secondaryButtonProps` and `buttonsAlignment`, with the component owning ordering and variants: https://github.com/MetaMask/metamask-design-system/blob/main/packages/design-system-react-native/src/components/BottomSheetFooter/BottomSheetFooter.types.ts. This web version currently splits responsibility between top-level `onSubmit` / `onCancel` and nested button prop bags, which makes the cross-platform model harder to share in docs/tests and leaves more room for divergence. Since this component is new, I think it would be cleaner to adopt the same shape here and let button presence be driven by `primaryButtonProps` / `secondaryButtonProps`.

packages/design-system-react/src/components/ModalFooter/ModalFooter.tsx
```diff
@@ -0,0 +1,70 @@
...
+        flexWrap={BoxFlexWrap.Wrap}
+        alignItems={BoxAlignItems.Center}
+        gap={4}
+        {...containerProps}
+        className={twMerge(
+          'mx-auto flex max-w-[360px]',
+          containerProps?.className,
+        )}
+      >
+        {onCancel && (
```
georgewrmarshall: suggestion: If we align this with RN `BottomSheetFooter`, I think the component should own the button semantics the same way the native implementation does: render secondary/primary from `secondaryButtonProps` / `primaryButtonProps`, enforce `ButtonVariant.Secondary` / `ButtonVariant.Primary` internally, and make layout explicit through a `buttonsAlignment` prop instead of hardcoding the current wrapped-horizontal arrangement. RN already does that here: https://github.com/MetaMask/metamask-design-system/blob/main/packages/design-system-react-native/src/components/BottomSheetFooter/BottomSheetFooter.tsx. That would also avoid exposing `variant` / `onClick` as something consumers can use to partially bypass the footer contract.

packages/design-system-react/src/components/ModalFooter/ModalFooter.types.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  cancelButtonProps?: ButtonPropsWithDataAttrs;
+  /**
+   * Optional props passed to the inner `Box` that wraps the submit/cancel
+   * buttons. Use this to override the default max-width, background, or any
+   * other Box utility prop. The default max-width of 360px is applied via
+   * className and will be overridden if `containerProps.className` provides
+   * its own `max-w-*` utility (`twMerge` lets the consumer's class win).
+   */
+  containerProps?: Omit<BoxProps, 'children'> & {
```
georgewrmarshall: suggestion: Can we exclude `asChild` from `containerProps` here? `BoxProps` currently leaks `asChild`, but this footer’s internal `Box` can render two buttons. If a consumer passes `containerProps={{ asChild: true }}`, `Box` switches to Radix `Slot`, which only supports a single child element, so this public type surface can typecheck and then fail at runtime. `Omit<BoxProps, 'children' | 'asChild'>` seems closer to the intended contract.

georgewrmarshall (changes requested):
I compared this against the existing React Native `BottomSheetFooter` API as well as the extension `modal-footer` migration target. The migration direction looks reasonable, but I think we should align the new public footer API across platforms before this settles: the current web `ModalFooter` introduces a different action model (`onSubmit` / `onCancel` + button prop bags) and a fixed layout shape where RN already has a cleaner `primaryButtonProps` / `secondaryButtonProps` + `buttonsAlignment` contract.

georgewrmarshall (dismissed):
I rechecked the current head against the earlier review round. The footer API is now aligned with the React Native `BottomSheetFooter` direction: button presence is driven by `primaryButtonProps` / `secondaryButtonProps`, `buttonsAlignment` is part of the public contract, the component owns button variants, and the old `containerProps` surface is gone. I also ran the focused `ModalFooter` test file on the current head and it passes. I don’t have any remaining blocking findings on this round.

georgewrmarshall (approved):

PR #1141, opened 2026-04-30
feat: [DSRN] Added TextArea

brianacnguyen:
## **Description**

This PR adds **TextArea** to `@metamask/design-system-react-native`: a multiline field built from the existing **Input** inside a styled **Box** (border, focus/error/disabled states, min height, transparent inner input). It aligns with the shared field model by introducing **`TextAreaPropsShared`** in `@metamask/design-system-shared` (ADR-0004), mirroring `TextField` without start/end accessories.

**Input** was updated so multiline mode works correctly: typography uses the text variant + font classes instead of single-line input metrics, the inner `TextInput` receives `multiline`, and the iOS-only `lineHeight: 0` placeholder workaround is **skipped** when multiline so paragraph layout is not broken.

Also included: **README**, **Storybook** stories, and **unit tests** for TextArea; **Input** tests were extended for multiline behavior.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-573 

## **Manual testing steps**

1. From the repo root, run React Native Storybook (e.g. `yarn storybook:ios` or `yarn storybook:android`).
2. Open the **TextArea** stories and verify default, disabled, error, and focus border behavior; type multiple lines and confirm layout and placeholder alignment.
3. Spot-check a story or screen that still uses **Input** in single-line mode to confirm no regression.

## **Screenshots/Recordings**

_If applicable, add screenshots and/or recordings to visualize the before and after of your change._

### **After**

https://github.com/user-attachments/assets/155caac5-87fc-4f2e-b35d-5f360765ca02

---

> [!NOTE]
> **Medium Risk**
> Updates shared `Input` styling/placeholder behavior and introduces a new multiline `TextArea`, so regressions could affect existing input rendering across platforms (especially iOS/Android placeholder/layout differences). Scope is UI-focused with good test coverage, but touches a widely-used primitive.
> 
> **Overview**
> Adds a new `TextArea` component to `@metamask/design-system-react-native`: a multiline, bordered field wrapper around `Input` with focus/error/disabled chrome, `inputProps` passthrough, and separate refs for the root container vs. inner `TextInput`.
> 
> Updates `Input` to properly support multiline usage by forwarding `multiline`, using variant typography styles (instead of single-line metrics) for multiline, and scoping the iOS `lineHeight: 0` placeholder workaround to **single-line only**.
> 
> Extends/updates unit tests and Storybook to cover the new `TextArea` and the updated `Input` multiline + iOS placeholder behaviors, and aligns shared prop contracts by introducing `TextAreaPropsShared` and refactoring `TextFieldPropsShared` to build on `InputPropsShared` (which now includes `placeholder` and `autoFocus`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb78795917d209f59fb852f0b02ee64c1e625f5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/design-system-shared/src/types/TextArea/TextArea.types.ts
```diff
@@ -0,0 +1,11 @@
+import type { TextFieldPropsShared } from '../TextField/TextField.types';
+
+/**
+ * TextArea shared props (ADR-0004). Same field contract as `TextField` for
+ * multiline use, without start/end accessories. Platform packages layer
+ * container and `TextInput` behavior.
+ */
+export type TextAreaPropsShared = Omit<
```
georgewrmarshall, line 10: suggestion: Overall this PR looks close, but I think this should wait for PR 1043 and then derive from `InputPropsShared` instead of `TextFieldPropsShared`. `TextArea` reads as a multiline input rather than a `TextField` minus accessories, and PR 1043 establishes the shared `Input` contract (`value`, `textVariant`, `isDisabled`, `isReadOnly`, `isStateStylesDisabled`) that matches this component much more closely. Using `TextFieldPropsShared` here pulls in wrapper-specific concepts that don’t really belong in the textarea contract.

georgewrmarshall (changes requested):
Overall this PR looks good, and the RN implementation/tests are in solid shape. I think the remaining issue is mostly sequencing and shared-API shape: PR #1043 should land first so this component can build on `InputPropsShared`, and then `TextAreaPropsShared` can be based on the shared input contract rather than inheriting from `TextFieldPropsShared`.

packages/design-system-shared/src/types/TextArea/TextArea.types.ts
```diff
@@ -0,0 +1,27 @@
...
+ * `InputPropsShared` contract. Platform packages layer container and
+ * `TextInput` behavior on top.
+ */
+export type TextAreaPropsShared = Omit<
+  InputPropsShared,
+  'isStateStylesDisabled'
+> & {
+  /**
+   * Called when the text changes. Uses React Native `TextInput` naming; web
+   * implementations may map this from the native input change event.
```
georgewrmarshall: suggestion: `TextAreaPropsShared` is extending the existing `onChangeText` pattern in the shared layer, but `onChangeText` is still a React Native-specific callback name. Since this lives in `@metamask/design-system-shared`, can we avoid deepening that platform-specific contract here?

packages/design-system-shared/src/types/TextField/TextField.types.ts
```diff
@@ -1,42 +1,30 @@
...
  * TextField shared props (ADR-0004).
  *
- * Platform-independent fields for a controlled text field and optional chrome
- * (accessories, custom input slot). Styling, `testID`, native-only `inputProps`,
- * and typed focus/blur handlers stay on the platform layer.
+ * Layers controlled text-field chrome (accessories, custom input slot, error
+ * state, etc.) on top of the shared `InputPropsShared` contract. Styling,
+ * `testID`, native-only `inputProps`, and typed focus/blur handlers stay on
+ * the platform layer.
  */
```
georgewrmarshall, line 12: suggestion: `TextFieldPropsShared` now inherits from `InputPropsShared`, which means `textVariant` becomes part of the `TextField` public surface. The extension `TextField` API does not expose typography control, so was that expansion intentional? If not, I think `TextField` should passively inherit `textVariant` from its public surface and keep the internal `BodyMd` default so the migration target stays aligned with the source component.

packages/design-system-shared/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
georgewrmarshall: suggestion: I don't think `packages/design-system-shared/CHANGELOG.md` should be updated in this PR if changelog entries are handled by the release process. Can we drop this file from the changeset here?

georgewrmarshall (changes requested):
The RN implementation is close, and the earlier dependency/focus-state issues look resolved, but I think there are still a few API/process issues to address before merge. This PR extends the React Native-specific `onChangeText` callback name further into `@metamask/design-system-shared`, appears to expose `textVariant` on `TextField` through shared-type inheritance even though the source extension component does not expose typography control, and manually updates `packages/design-system-shared/CHANGELOG.md` even though changelog updates are part of the release process. I think those should be cleaned up before this lands.

packages/design-system-react-native/src/components/Input/Input.test.tsx
```diff
@@ -46,13 +46,14 @@ describe('Input', () => {
     expect(input).toHaveStyle({ opacity: tw`opacity-50`.opacity });
   });
 
-  it('applies iOS placeholder lineHeight workaround when placeholder is visible', () => {
-    if (Platform.OS !== 'ios') {
-      return;
-    }
-
+  it('applies iOS placeholder lineHeight workaround when placeholder is visible and multiline is false', () => {
```
georgewrmarshall: non-blocking: For this iOS-specific behavior, I think the test would be stronger if it explicitly stubbed `Platform.OS = 'ios'` and restored it afterward rather than relying on whatever platform the runner happens to report. The same applies to the negative-path test below for the non-iOS branch.

packages/design-system-react-native/src/components/Input/Input.tsx
```diff
@@ -82,8 +85,11 @@ export const Input = forwardRef<TextInput, InputProps>(
     );
```
georgewrmarshall, line 107: non-blocking: The multiline typography split looks intentional, but the current structure is pretty hard to read. As written, single-line input uses the input-metrics path while multiline switches to the full `text-*` styling path, and `fontClass` gets applied in different places depending on `multiline`. A small refactor here would make that distinction much easier to follow.

packages/design-system-react-native/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,119 @@
...
+      placeholder,
+      isReadOnly,
+      onBlur,
+      onFocus,
+      autoFocus = false,
+      inputProps,
+      inputRef,
+      isDisabled = false,
+      isError = false,
+      textVariant = TextVariant.BodyMd,
```
georgewrmarshall: non-blocking: Is this necessary as we already default to `TextVariant.BodyMd` via `Input`?

packages/design-system-react-native/src/components/Input/Input.tsx
```diff
@@ -59,7 +61,7 @@ export const Input = forwardRef<TextInput, InputProps>(
     const inputStyle = useMemo(
```
georgewrmarshall: non-blocking: `Input` still uses `border-primary-default` for its standalone focus styling, while `TextField` and `TextArea` now own focus chrome at the container level and use `border-default`. I do not think that is introduced by this PR, but it does look like field-stack inconsistency worth cleaning up in follow-up work.

georgewrmarshall (approved):
The current revision looks ready to merge. The earlier shared-type/API concerns appear resolved: TextAreaPropsShared now builds on InputPropsShared, TextField no longer exposes textVariant through shared inheritance, and the changelog edit is gone. The RN implementation and targeted multiline/input tests also look solid from this pass.

PR #1153, opened 2026-05-05
feat: `Popover` migration

kirillzyusko:
## **Description**

Added `Popover` component to DSR.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-313

## **Manual testing steps**

1. Open Storybook
2. Check `Popover` component

### **Before**

https://github.com/user-attachments/assets/1331895e-56fc-4778-9b20-99ebddc458f0

### **After**

https://github.com/user-attachments/assets/8f64a5a5-5e67-47e5-bb01-ab2d6ca41e91

---

> [!NOTE]
> **Medium Risk**
> Touches a widely reused overlay primitive with document-level listeners and positioning semantics; regressions could affect menus/tooltips across extension call sites, though behavior is heavily tested and documented.
> 
> **Overview**
> Adds a **`Popover`** component to `@metamask/design-system-react` for anchored overlays (menus, tooltips, dialogs), with **Storybook**, **Jest** coverage, package **exports**, and a large **`MIGRATION.md`** section for moving off the extension `component-library` popover.
> 
> Positioning is implemented with **`@floating-ui/react-dom`** (`useFloating`, flip/shift/autoPlacement, arrow, hide), while keeping the familiar prop contract (`referenceElement`, `isOpen`, placement, portal, match width, Esc / click-outside). The surface is a fixed **`div`** with tokenized **`Box`** styling and Tailwind (`rounded-lg`, `shadow-md`, reference-hidden `data-*` variants) instead of polymorphic `as`, Box utility passthrough, and `.mm-popover` SCSS. **`PopoverPosition`** / **`PopoverRole`** are const objects with string unions (ADR-0003); document listeners for Escape and outside click are registered and removed with matching **`{ capture: true }`** options.
> 
> Extension adopters are guided toward import-path-only migrations where possible, with **`className`** / **`arrowProps.className`** for overrides. Note: some in-repo comments and migration text still mention **`react-popper`**, while the runtime dependency is **Floating UI**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1b8b481a015b7ebc096b9d3615eebe8cf80ee9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kirillzyusko:
Socket Security detected that the npm publisher changed from ryan.roemer (a personal account) to formidablelabs (the org account). Socket flags all publisher changes because account takeovers are a real supply chain attack vector.

**Why it's safe to dismiss:** The transfer is a legitimate housekeeping move within the same organization:

- The GitHub repo has always been at github.com/FormidableLabs/react-fast-compare — the org always owned the code
- ryan.roemer is the co-founder of Formidable Labs, which was acquired by NearForm in Oct 2023
- Moving the npm publish right from the founder's personal account to the org account is expected after an acquisition
- react-popper is also a Formidable Labs project, so the same org controls both the dep and its consumer
- 3.2.2 is the latest version (published ~3 years ago) — no new code was injected

Recommended action: Add a @SocketSecurity ignore comment on the PR to dismiss the alert.

@georgewrmarshall any thoughts on this? 👀

packages/design-system-react/package.json
```diff
@@ -49,9 +49,11 @@
   "dependencies": {
     "@metamask/design-system-shared": "workspace:^",
     "@metamask/jazzicon": "^2.0.0",
+    "@popperjs/core": "^2.4.0",
     "@radix-ui/react-slot": "^1.1.0",
     "blo": "^2.0.0",
     "react-focus-lock": "^2.9.4",
+    "react-popper": "^2.2.3",
```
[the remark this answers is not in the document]
georgewrmarshall: This is interesting I wonder if we should migrate to https://floating-ui.com/docs/?

georgewrmarshall: non-blocking: `react-popper@2.3.0` only declares React peers up to 18, but `@metamask/design-system-react` advertises React 19 support. That creates a compatibility mismatch for consumers even if the component itself works. If we keep this dependency in this PR, I think we should at least track a follow-up to move to `@floating-ui/react`, which currently supports `react >=17` / `react-dom >=17`.

packages/design-system-react/src/components/Popover/Popover.types.ts
```diff
@@ -0,0 +1,131 @@
...
+ * Use `Dialog` when the contents are interactive, `Tooltip` for purely
+ * informational content.
+ */
+export const PopoverRole = {
+  Tooltip: 'tooltip',
+  Dialog: 'dialog',
+} as const;
+export type PopoverRole = (typeof PopoverRole)[keyof typeof PopoverRole];
+
+export type PopoverProps = Omit<ComponentProps<'div'>, 'role'> & {
```
georgewrmarshall, line 37: question: this migration currently keeps `Popover` types local to `design-system-react`, and I’m not seeing the shared-type / React Native side of the migration. Our migration docs and ADR-0004 patterns usually expect migrated components to land with shared types and cross-platform readiness, even if alignment is partial. Is this intended as an explicit React-only exception, or should that follow-up be captured somewhere?

georgewrmarshall (commented):
## Component Migration Review: Popover

Reviewed against:
- Component migration strategy
- `.cursor/rules/component-migration.md`
- `.cursor/rules/component-architecture.md`
- `.cursor/rules/styling.md`

### Summary
The Popover migration looks close. The implementation, stories, tests, and migration docs are in good shape overall, but I left a few follow-ups around the positioning dependency choice, migration architecture consistency, and styling conventions in the new files.

### Key Findings
- `react-popper` introduces a React support mismatch against this package’s advertised React 19 peer range.
- This migration currently lands as React-only with local types, which doesn’t fully match the documented migration workflow for shared types and cross-platform readiness.
- A couple of new files lean on static inline/CSS object styling where our styling guidance would prefer Tailwind classes or component props.

### Follow-up
Separate from this PR, it’s probably worth discussing a move from `react-popper` to `@floating-ui/react`. Floating UI’s current React package supports `react >=17` / `react-dom >=17`, while `react-popper` is in maintenance mode.

packages/design-system-react/src/components/Popover/README.mdx
```diff
@@ -0,0 +1,366 @@
...
+  <thead>
+    <tr>
+      <th align="left">TYPE</th>
+      <th align="left">REQUIRED</th>
+      <th align="left">DEFAULT</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td align="left"><code>ComponentProps&lt;'div'&gt;</code></td>
```
georgewrmarshall: non-blocking: the docs say `arrowProps` is `ComponentProps<'div'>`, but the actual public type is narrower: `Omit<ComponentProps<'div'>, 'ref' | 'style' | 'children'>`. I’d update the README (and MIGRATION guide) so consumers don’t think those props are supported overrides.

packages/design-system-react/src/components/Popover/Popover.constants.ts
```diff
@@ -0,0 +1,67 @@
...
+    container: { left: -20 },
+    visual: { transform: 'rotate(-45deg)' },
+  },
+};
+
+/**
+ * Static styles for the outer arrow container (sized to match `react-popper`'s
+ * default arrow modifier). Visibility is hidden — only the inner visual draws.
+ */
+export const POPOVER_ARROW_CONTAINER_STYLE: CSSProperties = {
```
georgewrmarshall: suggestion: can we move the static arrow styling pieces to Tailwind classes or Box props and leave only the truly dynamic placement/transform bits in `style`? The runtime positioning data makes some inline styles unavoidable here, but the fixed width/height/visibility/border pieces feel like they should follow our normal styling conventions more closely.

packages/design-system-react/src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+}) => (
+  <Box
+    ref={(node) => setRef(node)}
+    backgroundColor={BoxBackgroundColor.PrimaryDefault}
+    alignItems={BoxAlignItems.Center}
+    justifyContent={BoxJustifyContent.Center}
+    flexDirection={BoxFlexDirection.Row}
+    padding={4}
+    asChild
+    style={{ width: 200, height: 200 }}
```
georgewrmarshall: non-blocking: these fixed story dimensions would be better as Tailwind classes instead of inline styles. I noticed the same pattern in a few places in this file (`width` / `height` / `minWidth` / `minHeight`), and converting those would keep the stories aligned with the repo’s styling guidance.

georgewrmarshall:
Could we make sure to include these changes from the extension version https://github.com/MetaMask/metamask-extension/pull/42732 🙏

packages/design-system-react/src/components/Popover/Popover.types.ts
kirillzyusko: @georgewrmarshall I think it's intentional, because we don't have `Popover` equivalent on mobile side? And if we try to move something into shared folder it can be harder to re-work the code later, because we will think that we had an API design and intentionally made this split...

packages/design-system-react/package.json
kirillzyusko: Migrated to `floating-ui`!

packages/design-system-react/package.json
```diff
@@ -47,6 +47,7 @@
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch"
   },
   "dependencies": {
+    "@floating-ui/react-dom": "^2.1.2",
```
georgewrmarshall, line 50: non-blocking: Good catch adding @floating-ui/react-dom. The MIGRATION.md and README.mdx files still reference react-popper in several places and should be updated to reflect that this implementation uses Floating UI, not Popper.js.

packages/design-system-react/MIGRATION.md
```diff
@@ -2653,6 +2654,157 @@ For typical call sites — for example `ui/components/multichain/network-list-me
 - `ModalOverlay` no longer composes Box's polymorphic API. It always renders a `<div>` and forwards arbitrary HTML attributes (`id`, `role`, `data-*`, `aria-*`, `ref`) to it.
 - One-off styling that previously used Box utility props (e.g. `backgroundColor={BackgroundColor.overlayAlternative}`) should now use `className` with the equivalent Tailwind utility (e.g. `className="bg-overlay-alternative"`).
 
+### Popover Component
+
+The extension `popover` component maps to `Popover` in the design system. The runtime contract — `referenceElement`, `isOpen`, `position`, `role`, `hasArrow`, `matchWidth`, `flip`, `preventOverflow`, `referenceHidden`, `offset`, `isPortal`, `arrowProps`, `onPressEscKey`, `onClickOutside` — is preserved 1:1, including the `PopoverPosition` and `PopoverRole` value strings. The breaking changes are limited to the surrounding API surface: `react-popper` is now an internal runtime dependency (not a peer), the polymorphic `as` / Box style-utility passthrough is gone, and the legacy SCSS class hooks (`.mm-popover`, `.mm-popover__arrow`, `.mm-popover--reference-hidden`) are replaced by Tailwind utilities and inline styles.
```
georgewrmarshall, line 2745: non-blocking: This says `react-popper is now an internal runtime dependency` but the implementation uses @floating-ui/react-dom. All references to react-popper in this migration guide should be updated to @floating-ui/react-dom for accuracy.

packages/design-system-react/MIGRATION.md
```diff
@@ -2653,6 +2654,157 @@ For typical call sites — for example `ui/components/multichain/network-list-me
...
+| `role?: PopoverRole`                                                                                       | `role?: PopoverRole`                                                       | unchanged   | same string values; default `PopoverRole.Tooltip`                                                                                                                                                                 |
+| `hasArrow?: boolean`                                                                                       | `hasArrow?: boolean`                                                       | unchanged   | default `false`; the rendered notch rotates with the resolved placement                                                                                                                                           |
```
georgewrmarshall, line 2767: non-blocking: Same react-popper vs @floating-ui/react-dom discrepancy here. The positioning library column should say Floating UI, not Popper.js.

packages/design-system-react/MIGRATION.md
```diff
@@ -2653,6 +2654,157 @@ For typical call sites — for example `ui/components/multichain/network-list-me
...
+| `flip?: boolean`                                                                                           | `flip?: boolean`                                                           | unchanged   | forced on when `position === PopoverPosition.Auto`                                                                                                                                                                |
+| `referenceHidden?: boolean`                                                                                | `referenceHidden?: boolean`                                                | unchanged   | default `true`; behavior preserved (hides popover when `data-popper-reference-hidden="true"`) — implemented via Tailwind `data-[]:` variants instead of SCSS                                                      |
```
georgewrmarshall, line 2772: suggestion: Consider adding inline code examples or story links under each prop section in the README, similar to how the extension component-library documents props. This would improve discoverability for consumers migrating from the extension.

georgewrmarshall (approved):
Strong migration work with comprehensive documentation. The implementation correctly uses Floating UI and maintains full prop parity with the extension version. Approving with a few non-blocking documentation clarifications needed to align with the actual implementation.

Can we add code examples similar to the extension storybook. Looks like we are missing some stories as well can be added as follow up if easier


https://github.com/user-attachments/assets/170daa35-d0e5-4b8b-b332-8904e3539c91

PR #1165, opened 2026-05-07
chore: Align React Native and Expo dependencies with mobile

georgewrmarshall:
## **Description**

Aligns the React Native packages and Storybook app with the dependency baseline introduced in [MetaMask Mobile #29195](https://github.com/MetaMask/metamask-mobile/pull/29195) — the React Native 0.76.9 → 0.81.5 / React 18 → 19 / Expo 52 → 54 upgrade that mobile recently merged.

As part of this upgrade, the Storybook app scripts have been switched from native builds (`expo run:ios` / `expo run:android`) back to **Expo Go** (`expo start --ios` / `expo start --android`), which no longer require a full native build and are significantly faster to set up for local development.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-744

### `apps/storybook-react-native` — production dependencies

| Package | Before | After | Mobile (MetaMask/metamask-mobile#29195) |
|---|---|---|---|
| `react` | `18.3.1` | `19.1.0` | `19.1.0` ✅ |
| `react-native` | `0.76.9` | `0.81.5` | `0.81.5` ✅ |
| `expo` | `~52.0.49` | `~54.0.34` | `54.0.33` ✅ |
| `expo-asset` | `~11.0.5` | `~12.0.12` | `~12.0.12` ✅ |
| `expo-font` | `~13.0.4` | `~14.0.11` | `~14.0.11` ✅ |
| `expo-status-bar` | `~2.0.0` | `~3.0.9` | — |
| `react-native-gesture-handler` | `~2.25.0` | `~2.28.0` | `~2.28.0` ✅ |
| `react-native-reanimated` | `~3.17.2` | `~4.1.1` | `3.19.0` (mobile still compatible) |
| `react-native-worklets` | — | `0.5.1` | — |

### `apps/storybook-react-native` — dev dependencies

| Package | Before | After | Mobile (MetaMask/metamask-mobile#29195) |
|---|---|---|---|
| `@react-native-async-storage/async-storage` | `^1.23.1` | `2.2.0` | `2.2.0` ✅ |
| `@react-native-community/datetimepicker` | `^8.5.1` | `^8.5.1` | `^8.5.1` ✅ |
| `@react-native-community/slider` | `^4.5.6` | `5.0.1` | `5.0.1` ✅ |
| `@storybook/react-native` | `^10` | `10.4.0` | — |
| `@storybook/addon-ondevice-actions` | `^10` | `10.4.0` | — |
| `@storybook/addon-ondevice-backgrounds` | `^10` | `10.4.0` | — |
| `@storybook/addon-ondevice-controls` | `^10` | `10.4.0` (patched) | — |
| `@storybook/addon-ondevice-notes` | `^10` | `10.4.0` | — |
| `storybook` | `^10.3.1` | `10.3.6` | — |
| `@types/react` | `^18.2.0` | `^19.1.0` | `^19.1.0` ✅ |
| `@types/react-dom` | `^18.2.0` | `^19.1.0` | — |
| `react-dom` | `18.3.1` | `19.1.0` | — |
| `react-native-safe-area-context` | `^5.4.0` | `~5.6.0` | `~5.6.0` ✅ |
| `react-native-svg` | `~15.11.2` | `15.12.1` | `15.12.1` ✅ |
| `react-native-svg-transformer` | `^1.5.0` | `^1.5.3` | — |

### `packages/design-system-react-native` — dev dependencies

| Package | Before | After | Mobile (MetaMask/metamask-mobile#29195) |
|---|---|---|---|
| `react` | `18.3.1` | `19.1.0` | `19.1.0` ✅ |
| `react-native` | `0.76.9` | `0.81.5` | `0.81.5` ✅ |
| `react-test-renderer` | `^18.3.1` | `19.1.0` | `19.1.0` ✅ |
| `@react-native/babel-preset` | `0.76.9` | `0.81.5` | `0.76.9` |
| `@react-native/typescript-config` | `0.76.9` | `0.81.5` | `0.76.9` |
| `@types/react` | `^18.2.0` | `^19.1.0` | `^19.1.0` ✅ |
| `@types/react-test-renderer` | `^18` | `^19.1.0` | — |
| `@testing-library/react-hooks` | `^8.0.1` | removed | — |
| `@storybook/react-native` | `^10` | `10.4.0` | — |
| `react-native-gesture-handler` | `~2.25.0` | `~2.28.0` | `~2.28.0` ✅ |
| `react-native-reanimated` | `~3.17.2` | `~4.1.1` | `3.19.0` |
| `react-native-safe-area-context` | `^5.4.0` | `~5.6.0` | `~5.6.0` ✅ |
| `react-native-svg` | `~15.11.2` | `15.12.1` | `15.12.1` ✅ |
| `react-native-svg-transformer` | `^1.5.0` | `^1.5.3` | — |
| `react-native-worklets` | — | `0.5.1` | — |

### `packages/design-system-shared` — dev dependencies

| Package | Before | After | Mobile (MetaMask/metamask-mobile#29195) |
|---|---|---|---|
| `react` | `18.3.1` | `19.1.0` | `19.1.0` ✅ |
| `@types/react` | `^18.2.0` | `^19.1.0` | `^19.1.0` ✅ |

### `packages/design-system-twrnc-preset` — dev dependencies

| Package | Before | After | Mobile (MetaMask/metamask-mobile#29195) |
|---|---|---|---|
| `react` | `18.3.1` | `19.1.0` | `19.1.0` ✅ |
| `react-test-renderer` | `^18.3.1` | `19.1.0` | `19.1.0` ✅ |
| `@types/react` | — | `^19.1.0` | `^19.1.0` ✅ |
| `@types/react-test-renderer` | `^18` | `^19.1.0` | — |

### `apps/storybook-react` — dev dependencies

| Package | Before | After |
|---|---|---|
| `react` | `18.3.1` | `19.1.0` |
| `react-dom` | `18.3.1` | `19.1.0` |
| `@types/react` | `^18.2.0` | `^19.1.0` |
| `@types/react-dom` | `^18.2.0` | `^19.1.0` |
| `storybook` | `^10.3.1` | `10.3.6` |

## **Additional changes**

- Migrated `BottomSheetDialog` from the deprecated RNGH v1 `PanGestureHandler`/`useAnimatedGestureHandler` API to the RNGH v2 `GestureDetector`/`Gesture.Pan()` API
- Migrated hook tests from `@testing-library/react-hooks` (removed) to `renderHook` from `@testing-library/react-native`
- Added `react-native-worklets` to Jest `transformIgnorePatterns`
- Added a yarn patch for `@storybook/addon-ondevice-controls@10.4.0` (upstream bug fix)
- Switched Storybook scripts from `expo run:ios/android` (native builds) to `expo start --ios/android` (Expo Go) — previously required a full native build; now works out of the box with the upgraded Expo 54 baseline
- Pinned Storybook packages to exact versions for reproducibility

## **Manual testing steps**

1. Run `yarn storybook:ios` — confirm Expo Go launches and stories render
2. Run `yarn storybook:android` — confirm Expo Go launches and stories render
3. Run `yarn test` — confirm all tests pass
4. Run `yarn build` — confirm all packages build without errors
5. Run `yarn lint` — confirm no lint errors

### **Before**

Storybook running from native build because ExpoGo was broken in those versions of RN/Storybook

### **After**

Storybook running from ExpoGo in iOS and Android

https://github.com/user-attachments/assets/abd8fb84-539e-4c4a-b27f-ceb6c1cf5ed1

#### Preview builds

Testing MMDS `BottomSheet` with reanimated library updates in mobile

https://github.com/user-attachments/assets/7ad3dfe3-2557-4659-9757-4545857813cd

Extension running as expected

https://github.com/user-attachments/assets/e28d7626-9949-42af-9380-40c3b05a0258

- [x] I've applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-extension/blob/develop/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to broad React/React Native/Expo and Storybook version bumps plus a migration of `BottomSheetDialog` gesture handling to the RNGH v2 API, which can affect runtime interactions across devices.
> 
> **Overview**
> Aligns the monorepo’s Storybook + React Native toolchain with a newer Mobile baseline by upgrading **Expo 54, React 19, React Native 0.81** and related RN libs (including adding `react-native-worklets` and updating Jest transforms).
> 
> Migrates `BottomSheetDialog` from deprecated RNGH v1 `PanGestureHandler`/`useAnimatedGestureHandler` to RNGH v2 `GestureDetector`/`Gesture.Pan()`, and removes the `panGestureHandlerProps` pass-through from both `BottomSheet`/`BottomSheetDialog` (with accompanying story/test updates and migration docs).
> 
> Updates tests for React 19 by consolidating hook testing into `@testing-library/react-native`’s `renderHook`, adding a shared `createRenderer()` helper to wrap `react-test-renderer` in `act()`, and adjusting a few component typings/prop forwarding for updated React types.
> 
> Fixes Storybook RN breakage by switching config to `deviceAddons`, changing Storybook RN scripts to `expo start` (Expo Go), pinning Storybook packages, and applying a Yarn patch to `@storybook/addon-ondevice-controls` to add explicit `.js` extensions in built imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a54c916be78597dafde80e0a9777918dc32b10a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.yarn/patches/@storybook-addon-ondevice-controls-npm-10.4.0-dee6ebd17d.patch
```diff
@@ -0,0 +1,260 @@
...
+--- a/dist/ControlsPanel.js
++++ b/dist/ControlsPanel.js
+@@ -1,9 +1,9 @@
+ import { jsx as _jsx } from "react/jsx-runtime";
+ import { includeConditionalArg, } from 'storybook/internal/csf';
+ import React, { useCallback, useState } from 'react';
+-import NoControlsWarning from './NoControlsWarning';
+-import PropForm from './PropForm';
+-import { useArgs } from './hooks';
++import NoControlsWarning from './NoControlsWarning.js';
```
georgewrmarshall, line 12: The entire patch adds explicit `.js` file extensions to every internal import across the compiled `dist/` files. Metro bundler (the React Native bundler used by Expo) requires ESM imports to include the full file extension — without them it cannot resolve sibling modules. This is an upstream packaging bug in `@storybook/addon-ondevice-controls@10.4.0` that has not yet been released as a fix, so a yarn patch is the correct approach here. See the [yarn patch documentation](https://yarnpkg.com/cli/patch) for how this works.

This patch can be removed once `@storybook/addon-ondevice-controls` releases a version that includes this fix and we upgrade to it. Track upstream progress at the [storybook-react-native releases page](https://github.com/storybookjs/react-native/releases). When upgrading, check whether the new version ships corrected ESM imports — if it does, remove both this patch file and the `patch:` protocol prefix from the version specifier in `apps/storybook-react-native/package.json`.

apps/storybook-react-native/.rnstorybook/main.ts
```diff
@@ -5,7 +5,7 @@ const config: StorybookConfig = {
     '../../../packages/design-system-react-native/src/**/*.stories.@(js|jsx|ts|tsx)',
     '../stories/**/*.stories.@(js|jsx|ts|tsx)',
   ],
-  addons: [
+  deviceAddons: [
```
georgewrmarshall, line 8: `deviceAddons` is the Storybook 10 rename of `addons` in the React Native config. The previous key name `addons` no longer works in `@storybook/react-native@10` — using it silently results in no addons loading on-device. See the [Storybook React Native migration guide](https://github.com/storybookjs/react-native/blob/next/MIGRATION.md) for the full list of config key renames in v10.

apps/storybook-react-native/package.json
```diff
@@ -6,44 +6,45 @@
   "main": ".rnstorybook/index.ts",
   "files": [],
   "scripts": {
-    "android": "yarn prestorybook && expo run:android",
-    "ios": "yarn prestorybook && expo run:ios",
+    "android": "yarn prestorybook && expo start --android",
```
georgewrmarshall, line 9: Scripts are switched from `expo run:android/ios` to `expo start --android/ios` to target Expo Go instead of a custom dev client build. `expo run:` compiles and installs a full native build, which requires Xcode/Android Studio and takes significantly longer. `expo start` streams the JS bundle to the Expo Go app, enabling faster iteration without native rebuilds. This follows the pattern from [expo-template-storybook](https://github.com/dannyhw/expo-template-storybook). The `start` script also now runs `prestorybook` to ensure story discovery runs before launching, which the dev-client flow previously did not enforce.

apps/storybook-react-native/package.json
```diff
@@ -6,44 +6,45 @@
...
-    "react-native-reanimated": "~3.17.2"
+    "expo": "~54.0.34",
+    "expo-asset": "~12.0.12",
+    "expo-font": "~14.0.11",
+    "expo-status-bar": "~3.0.9",
+    "react": "19.1.0",
+    "react-native": "0.81.5",
+    "react-native-gesture-handler": "~2.28.0",
+    "react-native-reanimated": "~4.1.1",
+    "react-native-worklets": "0.5.1"
```
georgewrmarshall, line 28: `react-native-worklets` is a new required peer of `react-native-reanimated@~4.1.1`. Reanimated 4 extracted its worklet runtime into a separate package to allow it to be used independently. Without this package, Reanimated 4 will throw at runtime when any animated component is first rendered. It is listed here as a runtime dependency of the Storybook app because `BottomSheetDialog` and other animated components in the design system depend on it transitively.

apps/storybook-react-native/package.json
```diff
@@ -6,44 +6,45 @@
...
-    "react-native-safe-area-context": "^5.4.0",
-    "react-native-svg": "~15.11.2",
-    "react-native-svg-transformer": "^1.5.0",
-    "storybook": "^10.3.1",
+    "@react-native-async-storage/async-storage": "2.2.0",
+    "@react-native-community/datetimepicker": "8.4.4",
+    "@react-native-community/slider": "5.0.1",
+    "@storybook/addon-ondevice-actions": "10.4.0",
+    "@storybook/addon-ondevice-backgrounds": "10.4.0",
+    "@storybook/addon-ondevice-controls": "patch:@storybook/addon-ondevice-controls@npm%3A10.4.0#~/.yarn/patches/@storybook-addon-ondevice-controls-npm-10.4.0-dee6ebd17d.patch",
```
georgewrmarshall, line 38: The `patch:` protocol in the version specifier tells Yarn to apply the local patch file at `.yarn/patches/@storybook-addon-ondevice-controls-npm-10.4.0-dee6ebd17d.patch` on top of the published npm package after installation. All other Storybook on-device packages are pinned to exact versions (no `^` or `~`) to prevent unintended minor version drift — Storybook React Native has had breaking config changes between minor releases, so exact pinning ensures the patch remains applicable and the configuration stays consistent across installs.

packages/design-system-react-native/scripts/create-component/ComponentName/ComponentName.test.tsx
```diff
@@ -1,6 +1,5 @@
 import { useTailwind } from '@metamask/design-system-twrnc-preset';
-import { renderHook } from '@testing-library/react-hooks';
-import { render } from '@testing-library/react-native';
+import { renderHook, render } from '@testing-library/react-native';
```
georgewrmarshall, line 2: `renderHook` was moved into `@testing-library/react-native` directly in v12+ and the separate `@testing-library/react-hooks` package is no longer needed or maintained for React Native. Using two separate packages for `render` and `renderHook` caused version skew issues in practice. This template is the starting point for all new React Native components created via `yarn create-component:react-native`, so updating it here ensures every new component scaffold starts with the correct import pattern. See the [RTLN v12 migration guide](https://callstack.github.io/react-native-testing-library/docs/migration/v12) for details.

georgewrmarshall:
Addressing the Cursor Bugbot comment at #discussion_r3276645252 — this is a false positive on two counts.

First, `metro-react-native-babel-transformer` does not follow React Native's versioning scheme. The latest published version is `0.77.0` — there is no `0.81.x`. The version numbers are independent of the RN release cadence, so `0.76.9` staying at `0.76.9` is not a mismatch.

Second, the package was never actually used. The `react-native` jest preset sets `transform` to `babel-jest` directly (not this transformer), and the package's own jest config also overrides transform with `babel-jest`. Neither execution path ever loads `metro-react-native-babel-transformer` at runtime.

This was a stale devDependency inherited from an older RN setup. Removed in 0f2036c1.

apps/storybook-react/package.json
```diff
@@ -25,8 +25,8 @@
     "@testing-library/dom": "^9.0.0",
     "@testing-library/react": "^16.0.1",
     "@types/prop-types": "^15",
-    "@types/react": "^18.2.0",
-    "@types/react-dom": "^18.2.0",
+    "@types/react": "^19.1.0",
```
georgewrmarshall, line 28: These version bumps are required by the monorepo yarn constraints rule (expectConsistentDependenciesAndDevDependencies in yarn.config.cjs), which enforces that every workspace declaring the same dependency must use the same version range. Once react, react-dom, @types/react, and @types/react-dom were bumped to 19.x across the other packages in this PR, yarn constraints would fail here until this workspace matched. Running yarn constraints --fix resolves these automatically.

packages/design-system-react-native/package.json
```diff
@@ -61,30 +61,30 @@
...
     "metro-react-native-babel-transformer": "0.76.9",
-    "react": "18.3.1",
-    "react-native": "0.76.9",
-    "react-native-gesture-handler": "~2.25.0",
-    "react-native-reanimated": "~3.17.2",
-    "react-native-safe-area-context": "^5.4.0",
-    "react-native-svg": "~15.11.2",
-    "react-native-svg-transformer": "^1.5.0",
-    "react-test-renderer": "^18.3.1",
+    "react": "19.1.0",
```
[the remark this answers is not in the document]
georgewrmarshall: `metro-react-native-babel-transformer` has been removed

packages/design-system-react-native/src/components/ActionListItem/ActionListItem.test.tsx
```diff
@@ -269,18 +267,6 @@ describe('ActionListItem', () => {
       );
       expect(getByTestId('action-item')).toBeDefined();
     });
-
```
georgewrmarshall, line 272: Removed because the assertion — `expect(styleFn({ pressed: true })).toBeDefined()` — is a no-op: a style function always returns a defined value regardless of whether the correct class is applied. The test existed solely to hit the `pressed=true` branch for 100% coverage rather than to verify any observable behavior. Removing it intentionally drops branch coverage on this file; a per-file threshold exception has been added in `packages/design-system-react-native/jest.config.js` to document this. The pressed visual state is better verified on device via Storybook. See #1182 for the options considered and the path forward when upgrading to RNTL v14.

packages/design-system-react-native/src/test-utils/createRenderer.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import * as ReactTestRenderer from 'react-test-renderer';
+
+// Wraps ReactTestRenderer.create in act() as required by React 19.
+export const createRenderer = (element: React.ReactElement) => {
```
georgewrmarshall, line 5: React 19 throws a warning when `ReactTestRenderer.create()` is called outside of `act()`, so this wrapper centralises the fix rather than duplicating it across every test file that needs the renderer. It is excluded from coverage via `/test-utils/` in `jest.config.js` since it is infrastructure, not component logic. The renderer is used specifically in tests that need to inspect or invoke a `Pressable`'s `style` function directly — the only way to assert both branches of a pressed-state style in RNTL v13, where no interaction API can hold `pressed=true` during an assertion. See #1182 for the plan to migrate away from this pattern when RNTL v14 is stable.

packages/design-system-react-native/src/components/BadgeNetwork/BadgeNetwork.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
-import * as ReactTestRenderer from 'react-test-renderer';
 
+import { createRenderer } from '../../test-utils/createRenderer';
```
georgewrmarshall, line 3: The shared `createRenderer` utility replaces the inline `ReactTestRenderer.act() + create()` pattern that was duplicated across 7 test files. React 19 requires `ReactTestRenderer.create()` to be wrapped in `act()` — without it a warning is thrown. See `src/test-utils/createRenderer.tsx` for the implementation and #1182 for the longer-term plan to move away from `react-test-renderer` when RNTL v14 is stable.

georgewrmarshall:
Addressing Cursor Bugbot at #discussion_r3277501489 — this is a valid finding. The bugbot correctly identified that `simultaneousHandlers` was silently dropped.

Rather than patching `applyPanGestureProps` to add `gesture.simultaneousWith()`, we took the more decisive fix: `panGestureHandlerProps` has been **removed entirely** in e76b9d28. The prop was a compatibility shim from the RNGH v1 `PanGestureHandler` JSX API and was not used anywhere in the MetaMask Mobile or Extension consumer codebases. Since `simultaneousHandlers` was already broken and the entire prop surface was unverified against the v2 API, removal is cleaner than an incomplete fix.

The `MIGRATION.md` documents the breaking change and notes that first-class simultaneous gesture handling will be addressed as a follow-up if needed.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.tsx
```diff
@@ -19,13 +19,9 @@ import {
   Platform,
   KeyboardAvoidingView,
 } from 'react-native';
-import {
-  PanGestureHandler,
-  PanGestureHandlerGestureEvent,
-} from 'react-native-gesture-handler';
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
```
georgewrmarshall, line 22: The import swap from `PanGestureHandler` / `PanGestureHandlerGestureEvent` to `Gesture` / `GestureDetector` reflects the mandatory RNGH v1 → v2 API migration. `PanGestureHandler` was a JSX wrapper component; the v2 API uses a `GestureDetector` wrapper with a gesture object built via the fluent `Gesture.Pan()` builder. `useAnimatedGestureHandler` (which handled the event callbacks in Reanimated 3) was removed entirely from Reanimated 4, making this migration necessary.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.tsx
```diff
@@ -82,6 +77,7 @@ export const BottomSheetDialog = forwardRef<
     const currentYOffset = useSharedValue(screenHeight);
     const topOfDialogYValue = useSharedValue(0);
     const bottomOfDialogYValue = useSharedValue(screenHeight);
+    const gestureStartYOffset = useSharedValue(0);
```
georgewrmarshall, line 80: `gestureStartYOffset` replaces the `ctx` (context) object that `useAnimatedGestureHandler` provided for sharing state between `onStart`, `onActive`, and `onEnd` callbacks. In Reanimated 3, `ctx.startY` was the idiomatic way to pass the gesture origin across callbacks on the UI thread. In Reanimated 4, `useAnimatedGestureHandler` is gone and shared values are the correct replacement — they run on the UI thread the same way, just declared explicitly rather than managed implicitly by the hook.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.tsx
```diff
@@ -108,74 +104,82 @@ export const BottomSheetDialog = forwardRef<
...
-        const hasReachedDismissOffset =
-          latestOffset >
-          bottomOfDialogYValue.value *
-            DEFAULT_BOTTOMSHEETDIALOG_DISMISSTHRESHOLD;
-        // Check if the gesture's vertical speed has reached the threshold to determine a swipe action
-        const hasReachedSwipeThreshold =
-          Math.abs(velocityY) >
-          DEFAULT_BOTTOMSHEETDIALOG_SWIPETHRESHOLD_DURATION;
-        const isQuickDismissing = velocityY > 0;
+    const gestureHandler = useMemo(() => {
```
georgewrmarshall, line 107: The gesture is built inside `useMemo` rather than with a dedicated hook because RNGH v2 gestures are plain objects — they have no special hook requirements. `useMemo` prevents recreating the gesture object on every render (which would break gesture state continuity) and correctly re-creates it when `isInteractable` or the close handler changes. The callback name change from `onActive` to `onUpdate` is an RNGH v2 rename; the logic inside is identical to the previous implementation.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.tsx
```diff
@@ -280,11 +284,7 @@ export const BottomSheetDialog = forwardRef<
         enabled={keyboardAvoidingViewEnabled}
         {...props}
       >
-        <PanGestureHandler
-          {...panGestureHandlerProps}
-          enabled={isInteractable}
-          onGestureEvent={gestureHandler}
-        >
+        <GestureDetector gesture={gestureHandler}>
```
georgewrmarshall, line 287: `GestureDetector` accepts the composed gesture object directly rather than wiring it via `onGestureEvent` as the old `PanGestureHandler` did. The `enabled` flag, previously a JSX prop on `PanGestureHandler`, is now configured via `.enabled(isInteractable)` in the gesture builder chain. The consumer-facing `panGestureHandlerProps` escape hatch has been removed as a breaking change — see MIGRATION.md and issue #1183 for the full context and the plan to replace it with a first-class `gestureModifier` prop or adopt `@gorhom/bottom-sheet`.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.types.ts
```diff
@@ -1,11 +1,5 @@
 // Third party dependencies.
 import type { ViewProps } from 'react-native';
-import type { PanGestureHandlerProps } from 'react-native-gesture-handler';
-
-type BottomSheetDialogPanGestureHandlerProps = Omit<
-  PanGestureHandlerProps,
-  'enabled' | 'onGestureEvent'
->;
```
georgewrmarshall, line 3: The `PanGestureHandlerProps` import and `BottomSheetDialogPanGestureHandlerProps` type alias are gone because `panGestureHandlerProps` has been removed from the public API. The old type was semantically wrong in any case — it was the v1 JSX component's props type being used to describe v2 gesture configuration. The prop was audited against the MetaMask Mobile and Extension consumer codebases and found to be unused at any call site. See MIGRATION.md for the breaking change documentation and issue #1183 for the follow-up plan.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.test.tsx
```diff
@@ -14,29 +13,38 @@ import type { BottomSheetDialogRef } from './BottomSheetDialog.types';
...
-};
 
 jest.mock('react-native-gesture-handler', () => ({
-  PanGestureHandler: ({
-    children,
-    ...props
-  }: {
-    children: React.ReactNode;
-  }) => {
+  GestureDetector: ({ children }: { children: React.ReactNode }) => {
```
georgewrmarshall, line 16: The mock stubs only the four methods the component actually calls: `enabled` (returns the gesture object so method chaining works), `onStart`, `onUpdate`, and `onEnd`. `GestureDetector` takes the gesture object directly rather than mounting a JSX component with its own props, so there is no component props surface to capture — just the gesture builder API.

packages/design-system-react-native/src/components/BottomSheet/BottomSheet.tsx
```diff
@@ -34,15 +34,14 @@ export const BottomSheet = forwardRef<BottomSheetRef, BottomSheetProps>(
...
-      panGestureHandlerProps,
       ...props
     },
     ref,
   ) => {
     const tw = useTailwind();
     const { bottom: screenBottomPadding } = useSafeAreaInsets();
     const { y: frameY } = useSafeAreaFrame();
-    const postCallback = useRef<BottomSheetPostCallback>();
+    const postCallback = useRef<BottomSheetPostCallback | undefined>(undefined);
```
georgewrmarshall, line 44: React 19 tightened the `useRef` type signature to require an explicit initial value. The previous `useRef<BottomSheetPostCallback>()` no longer satisfies the type without passing `undefined` explicitly.

packages/design-system-react-native/src/components/BottomSheet/BottomSheet.tsx
```diff
@@ -86,9 +85,12 @@ export const BottomSheet = forwardRef<BottomSheetRef, BottomSheetProps>(
         }
         return true;
       };
-      BackHandler.addEventListener('hardwareBackPress', hardwareBackPress);
+      const subscription = BackHandler.addEventListener(
```
georgewrmarshall, line 88: `BackHandler.addEventListener` now returns a subscription object whose `.remove()` method is the correct cleanup mechanism in RN 0.65+. The previous pattern of calling `BackHandler.removeEventListener` directly was deprecated and is removed in RN 0.81.

georgewrmarshall:
Note on `BottomSheet.stories.tsx` line 120: `listGestureRef` is still passed to the `ScrollView` but is currently disconnected from the bottom sheet pan gesture. It was previously wired via `panGestureHandlerProps.simultaneousHandlers` but that forwarding was silently broken. Kept as a placeholder — see #1183 for the plan to restore simultaneous gesture handling.

packages/design-system-react-native/MIGRATION.md
```diff
@@ -41,6 +43,35 @@ This guide provides detailed instructions for migrating your project from one ve
 
 ## Version Updates
 
+### From version 0.23.0 to 0.x.0
+
+#### Removed `panGestureHandlerProps` from `BottomSheetDialog` and `BottomSheet`
```
georgewrmarshall, line 52: This documents a breaking change in the next release. The prop was audited against MetaMask Mobile and found unused at every call site. The simultaneousHandlers case — the only documented real-world use — was already silently dropped by the previous implementation, so no working functionality is removed. Issue #1183 tracks the follow-up to restore a proper consumer escape hatch for pan gesture configuration.

packages/design-system-react-native/src/components/BottomSheet/BottomSheet.stories.tsx
```diff
@@ -135,10 +135,6 @@ const ScrollableListTemplate = (args: BottomSheetProps) => {
           goBack={goBack}
           shouldNavigateBack={false}
           onClose={goBack}
-          panGestureHandlerProps={{
```
georgewrmarshall, line 138: The simultaneousHandlers prop was passed here to coordinate the scroll and dismiss gestures on Android — without it, the pan gesture intercepts scroll attempts inside the sheet and starts dismissing instead. However this was already silently broken: applyPanGestureProps never mapped simultaneousHandlers to gesture.simultaneousWithExternalGesture() in the v2 API. The ScrollableList story will still render correctly but the Android gesture conflict is unresolved. See issue #1183 for the plan to fix this properly, either via a gestureModifier callback prop or adoption of @gorhom/bottom-sheet.

packages/design-system-react-native/src/components/BottomSheet/BottomSheet.test.tsx
```diff
@@ -14,7 +14,6 @@ jest.mock('@metamask/design-system-twrnc-preset', () => ({
 // Capture callbacks wired by BottomSheet to BottomSheetDialog
 let capturedDialogOnClose: ((hasPendingAction?: boolean) => void) | undefined;
 let capturedDialogOnOpen: ((hasPendingAction?: boolean) => void) | undefined;
-let capturedPanGestureHandlerProps: unknown;
```
georgewrmarshall, line 17: capturedPanGestureHandlerProps and the test that used it are removed alongside the prop itself. Notably the test was passing simultaneousHandlers through the prop and asserting it arrived at BottomSheetDialog — but since applyPanGestureProps never actually applied it to the gesture, the prop was a no-op at runtime even while the passthrough test was green.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.test.tsx
```diff
@@ -14,29 +13,38 @@ import type { BottomSheetDialogRef } from './BottomSheetDialog.types';
...
     // eslint-disable-next-line @typescript-eslint/no-require-imports
     const { View } = require('react-native');
-    panGestureHandlerPropsRef.current = props;
-    return <View {...props}>{children}</View>;
+    return <View>{children}</View>;
+  },
+  Gesture: {
+    Pan: () => {
```
georgewrmarshall, line 22: Gesture.Pan() returns a chainable builder object that mirrors the RNGH v2 fluent API. Each event method (onStart, onUpdate, onEnd) stores its callback in handlers and updates gestureCallbacksRef so gesture handler callbacks test suite can invoke them directly without simulating real touch events. enabled() is a no-op that returns the gesture object to keep the chaining pattern intact — the component calls .enabled(isInteractable).onStart(...).onUpdate(...).onEnd(...).

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.test.tsx
```diff
@@ -260,41 +260,6 @@ describe('BottomSheetDialog', () => {
...
-        <Text>Test Child</Text>
-      </BottomSheetDialog>,
-    );
-
-    expect(panGestureHandlerPropsRef.current.enabled).toBe(false);
-    expect(panGestureHandlerPropsRef.current.onGestureEvent).not.toBe(
-      externalOnGestureEvent,
-    );
-  });
-
```
georgewrmarshall, line 297: Removing test for removed `panGestureHandlerProps` prop

packages/design-system-react-native/src/components/Card/Card.types.ts
```diff
@@ -25,4 +25,4 @@ export type CardProps = {
     TouchableOpacityProps,
     'onPress' | 'style' | 'children'
   >;
-} & ViewProps;
+} & TouchableOpacityProps;
```
georgewrmarshall, line 28: Pragmatic workaround: using TouchableOpacityProps as the base type instead of ViewProps satisfies TypeScript stricter JSX spread checking in React 19 without an unsafe cast, since TouchableOpacityProps extends ViewProps and the View branch accepts the superset cleanly. The correct long-term fix is a discriminated union on onPress so consumers get precise types for each branch, but that complexity is not warranted for this component right now. This component is a candidate for revisiting.

packages/design-system-react-native/src/components/Card/Card.tsx
```diff
@@ -22,10 +22,10 @@ export const Card = ({
   if (onPress) {
     return (
       <TouchableOpacity
+        {...props}
```
georgewrmarshall, line 25: props is spread first so that onPress, style, and touchableOpacityProps all take precedence over anything in the generic passthrough. Previously props was last, meaning a consumer could accidentally override the component-controlled onPress or style. The cast is gone because CardProps now extends TouchableOpacityProps directly — see Card.types.ts.

packages/design-system-react-native/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -1,7 +1,5 @@
-/* eslint-disable @typescript-eslint/no-require-imports */
 import { useTailwind } from '@metamask/design-system-twrnc-preset';
-import { renderHook } from '@testing-library/react-hooks';
-import { render, fireEvent } from '@testing-library/react-native';
+import { fireEvent, render, renderHook } from '@testing-library/react-native';
 import type { ReactElement } from 'react';
```
georgewrmarshall, line 3: renderHook moved from the now-unmaintained @testing-library/react-hooks into @testing-library/react-native directly in v12. Consolidating to a single import also removes the eslint-disable comment that was suppressing the no-require-imports rule for the inline require calls this test previously used.

packages/design-system-react-native/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -181,21 +181,22 @@ describe('Checkbox', () => {
 
   it('applies pressed container styles', () => {
     const fn = jest.fn();
-    const rtr = require('react-test-renderer');
-    const RN = require('react-native');
-    const tree = rtr.create(
+    const tree = createRenderer(
```
georgewrmarshall, line 184: Three changes here. First, the inline require('react-test-renderer') and rtr.act() wrappers are replaced with the shared createRenderer utility, which handles the React 19 act() requirement in one place. Second, findByType(RN.Pressable) is replaced with findByProps({ accessibilityRole: 'checkbox' }) — querying by component type ties the test to the internal implementation, so if Checkbox ever wraps Pressable differently the test would break silently; querying by accessibilityRole tests the public contract instead. Third, the second createRenderer call renders the children function directly with pressed: true to reach the pressed style branch — the same pattern used across ButtonBase and other components, since RNTL v13 has no way to hold Pressable in a pressed state during an assertion.

packages/design-system-react-native/src/components/temp-components/Jazzicon/Jazzicon.test.tsx
```diff
@@ -51,7 +51,7 @@ describe('Jazzicon Component', () => {
           borderRadius: 0,
         },
       }),
-      expect.any(Object),
+      undefined,
```
georgewrmarshall, line 54: React 19 removed the legacy context as the second argument when calling component functions. In React 18, components were called as (props, legacyContext) where legacyContext was an object — so expect.any(Object) was the correct matcher. In React 19 that argument is undefined, making expect.any(Object) fail. This same change applies at lines 70, 92, and 108. It affects any test that asserts on mock component call arguments using toHaveBeenCalledWith.

packages/design-system-react-native/src/components/Toast/Toast.tsx
```diff
@@ -11,7 +11,7 @@ import type { RefObject } from 'react';
 import type { ToastOptions, ToastProps, ToastRef } from './Toast.types';
 import { ToastView } from './ToastView';
 
-let registeredRef: RefObject<ToastRef> | null = null;
+let registeredRef: RefObject<ToastRef | null> | null = null;
```
georgewrmarshall: React 19 tightened the RefObject type to explicitly include null — createRef<T>() now returns RefObject<T | null> rather than RefObject<T>, reflecting that ref.current is null before the component mounts. The outer null on registeredRef is separate and intentional: it tracks whether any Toast component has registered itself at all.

packages/design-system-react-native/src/components/Toast/Toast.test.tsx
```diff
@@ -86,7 +86,7 @@ jest.mock('../AvatarIcon', () => {
 
 // Helper to show toast and wait for render.
 const showToastAndWait = async (
-  toastRef: React.RefObject<ToastRef>,
+  toastRef: React.RefObject<ToastRef | null>,
```
georgewrmarshall: Mirrors the type change in Toast.tsx — React 19 RefObject now includes null in the type parameter. The test type annotation must match or TypeScript rejects the createRef<ToastRef>() return value being assigned to the variable.

packages/design-system-react-native/package.json
```diff
@@ -61,30 +61,29 @@
...
+    "react-native": "0.81.5",
+    "react-native-gesture-handler": "~2.28.0",
+    "react-native-reanimated": "~4.1.1",
+    "react-native-safe-area-context": "~5.6.0",
+    "react-native-svg": "15.12.1",
+    "react-native-svg-transformer": "^1.5.3",
+    "react-native-worklets": "0.5.1",
```
georgewrmarshall, line 85: react-native-worklets is a new required peer of react-native-reanimated 4.x. Reanimated 4 extracted its worklet runtime into a separate package. Without it, any component using Reanimated (including BottomSheetDialog) throws at runtime. react-test-renderer is pinned to an exact version rather than a range because it must match the React version exactly — a version mismatch causes cryptic hook errors in tests. @testing-library/react-hooks is removed since renderHook is now included directly in @testing-library/react-native v12+, and metro-react-native-babel-transformer was a stale unused devDependency.

packages/design-system-react/src/components/ModalFocus/ModalFocus.types.ts
```diff
@@ -8,12 +8,12 @@ export type ModalFocusProps = {
   /**
    * The `ref` of the element to receive focus initially.
    */
-  initialFocusRef?: RefObject<FocusableElement>;
+  initialFocusRef?: RefObject<FocusableElement | null>;
```
georgewrmarshall, line 11: Same React 19 RefObject type tightening seen in Toast.tsx and BottomSheet.tsx. createRef<T>() now returns RefObject<T | null> to accurately reflect that ref.current is null before the referenced element mounts.

packages/design-system-react/src/components/Text/Text.constants.ts
```diff
@@ -72,7 +72,7 @@ export const TWCLASSMAP_TEXT_VARIANT_FONTWEIGHT: Record<TextVariant, string> = {
 
 export const MAP_TEXT_VARIANT_TAG: Record<
   TextVariant,
-  keyof JSX.IntrinsicElements
+  keyof React.JSX.IntrinsicElements
```
georgewrmarshall, line 75: React 19 deprecated the global JSX namespace in favour of React.JSX. The global JSX namespace will be removed in a future release, so all references need to be qualified with React.JSX to stay forward-compatible.

packages/design-system-react/src/components/Text/Text.tsx
```diff
@@ -36,7 +36,9 @@ export const Text: React.FC<TextProps> = ({
 }) => {
   // When asChild is true, use Radix Slot to merge props onto the child component.
   // Otherwise, render the semantic HTML element mapped to this variant (e.g. h1-h4, p).
-  const Component = asChild ? Slot : MAP_TEXT_VARIANT_TAG[variant];
+  const Component: React.ElementType = asChild
```
georgewrmarshall, line 39: React 19 TypeScript support became stricter about using a variable as a JSX element type without an explicit annotation. Without React.ElementType, TypeScript cannot verify that Component is a valid JSX constructor and raises an error when it appears as <Component .../>. The runtime behaviour is unchanged.

package.json
```diff
@@ -62,6 +62,7 @@
   "resolutions": {
     "minimist": "1.2.8",
     "plist": "3.0.5",
+    "pretty-format/react-is": "^19.0.0",
```
georgewrmarshall, line 65: Jest's pretty-format package pulls in react-is as a peer dependency and defaults to the React 18 version. Without this resolution, pretty-format and the React 19 runtime disagree on how to identify React elements, causing cryptic test output formatting errors. The resolution forces all consumers of pretty-format/react-is to use the React 19-compatible version.

packages/design-system-react-native/src/components/ButtonBase/ButtonBase.test.tsx
```diff
@@ -1,20 +1,17 @@
...
 import { useTailwind } from '@metamask/design-system-twrnc-preset';
-import { renderHook } from '@testing-library/react-hooks';
-import { render, fireEvent } from '@testing-library/react-native';
+import { fireEvent, render, renderHook } from '@testing-library/react-native';
 import React from 'react';
 import { View, Text } from 'react-native';
-import * as ReactTestRenderer from 'react-test-renderer';
+import type { ReactTestInstance } from 'react-test-renderer';
+
+import { createRenderer } from '../../test-utils/createRenderer';
```
georgewrmarshall, line 8: Four import-level changes here. renderHook is consolidated from the unmaintained @testing-library/react-hooks into @testing-library/react-native (v12+). The import * as ReactTestRenderer namespace import is replaced with a type-only import of ReactTestInstance — the only thing still needed from react-test-renderer for typing the tree traversal results. createRenderer from the shared utility wraps ReactTestRenderer.create in act(), required by React 19. The beforeAll/module-level tw variable is replaced with a getTw() helper so each test that needs tw calls renderHook at point-of-use — this avoids stale state issues if the hook result is accessed before beforeAll completes in async test runners.

packages/design-system-react-native/jest.config.js
```diff
@@ -33,11 +33,18 @@ module.exports = merge(baseConfig, {
       lines: 87,
       statements: 87,
     },
+    // pressed && !isDisabled branch in getPressableStyle is not unit-testable without
+    // react-test-renderer internals (see https://github.com/MetaMask/metamask-design-system/issues/1182).
+    // Verified visually via Storybook on device.
+    './src/components/ActionListItem/ActionListItem.tsx': {
```
georgewrmarshall, line 39: The pressed && !isDisabled branch in ActionListItem's getPressableStyle is only reachable by directly invoking the Pressable style function with { pressed: true }, which requires react-test-renderer internals. The original test that covered this branch used toBeDefined() — an assertion guaranteed to pass regardless of implementation — so removing it loses no meaningful signal. The pressed visual state is verified on device via Storybook instead. See issue #1182 for the options considered and the plan to migrate away from this pattern when RNTL v14 is stable.

packages/design-system-react-native/jest.config.js
```diff
@@ -33,11 +33,18 @@ module.exports = merge(baseConfig, {
...
+      branches: 85,
+    },
   },
 
   // Add coverage ignore patterns
   coveragePathIgnorePatterns: [
     'index.ts',
+    '/test-utils/', // shared test utilities
```
georgewrmarshall, line 57: src/test-utils/ holds the shared createRenderer utility that wraps ReactTestRenderer.create in act() for React 19. It is test infrastructure, not component logic, so it is excluded from coverage the same way index.ts barrel files and constants files are.

packages/design-system-react-native/jest.config.js
```diff
@@ -54,7 +61,7 @@ module.exports = merge(baseConfig, {
     '^.+\\.(js|jsx|ts|tsx)$': 'babel-jest',
   },
   transformIgnorePatterns: [
-    'node_modules/(?!(react-native|@react-native|react-native-reanimated|@react-navigation|react-native-jazzicon|react-native-gesture-handler|react-native-safe-area-context)/)',
+    'node_modules/(?!(react-native|@react-native|react-native-reanimated|react-native-worklets|@react-navigation|react-native-jazzicon|react-native-gesture-handler|react-native-safe-area-context)/)',
```
georgewrmarshall, line 74: react-native-worklets is added here because Reanimated 4 extracted its worklet runtime into this separate package. Without it in transformIgnorePatterns, Jest tries to run the raw ESM source of react-native-worklets and throws a SyntaxError — the same reason all the other react-native-* packages are listed here.

packages/design-system-react-native/src/components/BottomSheetDialog/BottomSheetDialog.test.tsx
```diff
@@ -14,60 +12,45 @@ import type { BottomSheetDialogRef } from './BottomSheetDialog.types';
...
     // eslint-disable-next-line @typescript-eslint/no-require-imports
     const { View } = require('react-native');
-    panGestureHandlerPropsRef.current = props;
-    return <View {...props}>{children}</View>;
+    return <View>{children}</View>;
+  },
+  Gesture: {
+    Pan: () => {
+      const gesture: Record<string, unknown> = {};
+      const noop = () => gesture;
```
georgewrmarshall, line 24: The Gesture.Pan mock is now a minimal stub. The previous version captured onStart/onUpdate/onEnd callbacks into a gestureCallbacksRef so tests could invoke them directly, but every assertion in those tests was expect(handlers.onXxx).toBeDefined() — trivially true since the mock itself created those functions. The tests had descriptive names implying behavioral verification (clamping, threshold logic, dismissal) but produced no signal about whether that logic was correct. Removing them is honest: the gesture physics are covered by Storybook on device, and the coverage gap is documented in jest.config.js.

georgewrmarshall:
@metamaskbot publish-preview

georgewrmarshall:
@SocketSecurity ignore npm/@expo/metro-config@54.0.15
@SocketSecurity ignore npm/@expo/cli@54.0.24
@SocketSecurity ignore npm/@expo/devcert@1.2.1
@SocketSecurity ignore npm/@expo/plist@0.4.8
@SocketSecurity ignore npm/@expo/schema-utils@0.1.8
@SocketSecurity ignore npm/expo-keep-awake@15.0.8
@SocketSecurity ignore npm/expo-status-bar@3.0.9
@SocketSecurity ignore npm/lan-network@0.2.1
@SocketSecurity ignore npm/react-devtools-core@6.1.5
@SocketSecurity ignore npm/@react-native/debugger-frontend@0.81.5
@SocketSecurity ignore npm/@ungap/structured-clone@1.3.0
@SocketSecurity ignore npm/accepts@2.0.0
@SocketSecurity ignore npm/tar@7.5.13
@SocketSecurity ignore npm/yaml@2.8.4
@SocketSecurity ignore npm/@storybook/addon-ondevice-notes@10.4.0
@SocketSecurity ignore npm/@storybook/react-native@10.4.0

All are devDependencies in apps/storybook-react-native — not shipped in any published package or consumer bundle.

- @expo/* and expo-* packages (metro-config, cli, devcert, plist, schema-utils, keep-awake, status-bar): Official Expo team packages (https://github.com/expo/expo), part of the Expo 54 development toolchain.
- lan-network@0.2.1: Transitive dep of Metro bundler, used for local network interface detection in the Expo dev server.
- react-devtools-core@6.1.5: Official Meta/React DevTools package (https://github.com/facebook/react), required by the React Native debugger.
- @react-native/debugger-frontend@0.81.5: Official Meta package for the React Native debugger UI, versioned with RN 0.81.5.
- @ungap/structured-clone@1.3.0: Widely-used polyfill for structuredClone (https://github.com/ungap/structured-clone), transitive dep of Metro bundler.
- accepts@2.0.0: Standard HTTP content-negotiation utility (https://github.com/jshttp/accepts), transitive dep of the Express-based dev server in Metro.
- tar@7.5.13: Standard archive utility (https://github.com/isaacs/node-tar), used by npm/yarn tooling.
- yaml@2.8.4: Standard YAML parser (https://github.com/eemeli/yaml), transitive dep of various Expo config tools.
- @storybook/addon-ondevice-notes@10.4.0: Official Storybook React Native package (https://github.com/storybookjs/react-native). The obfuscation signal is a false positive common to bundled npm packages — Storybook minifies its dist output in a way that pattern-matches as obfuscated. Source is open and auditable.
- @storybook/react-native@10.4.0: Official Storybook React Native core package (https://github.com/storybookjs/react-native). The AI anomaly signal is low-confidence (62%) and typical of larger framework packages with complex module graphs.

brianacnguyen (approved):

PR #1170, opened 2026-05-11
feat: `TextField` migration (extension)

kirillzyusko:
## **Description**

Added `TextField` component to DSR. 

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-316

## **Manual testing steps**

1. Open Storybook app
2. Check `TextField` example

### **Before**

image

### **After**

image

---

> [!NOTE]
> **Low Risk**
> New UI component with tests and documentation only; no changes to auth, data handling, or existing consumer runtime behavior until extensions adopt the new import.
> 
> **Overview**
> Adds **`TextField`** to `@metamask/design-system-react` as the web implementation of the extension migration (DSYS-316): a **`forwardRef`** root `<div>` that wraps design-system **`Input`**, with Tailwind-based chrome (sizes, focus/error/disabled borders), optional **start/end accessories**, and **`inputElement`** / **`inputRef`** / **`inputProps`** for customization.
> 
> Exports **`TextField`**, **`TextFieldSize`**, **`TextFieldType`**, and **`TextFieldProps`** from the package barrel; web-only size/type const objects extend shared **`TextFieldPropsShared`**. Includes Storybook docs, stories, and unit tests. **`MIGRATION.md`** documents extension → design-system moves (`disabled`/`readOnly`/`error` → **`is*`**, **`InputComponent`** → **`inputElement`**, controlled-only, no polymorphic Box).
> 
> Shared package: comment tweak on **`TextFieldPropsShared`** and **`export { type ... }`** on the TextField types index.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbe0b8ae311f5d93ec3c3b3adabb816bf69c81fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/design-system-react/src/components/TextField/TextField.stories.tsx
+++ b/packages/design-system-react/src/components/TextField/TextField.stories.tsx
@@ -63,7 +63,24 @@ export const Default: Story = {
     placeholder: 'Sample placeholder',
     value: '',
   },
-  render: (args) => <ControlledTextField {...args} />,
+  render: (args) => {
+    const [value, setValue] = useState(args.value ?? '');
+
+    useEffect(() => {
+      setValue(args.value ?? '');
+    }, [args.value]);
+
+    return (
+      <TextField
+        {...args}
+        value={value}
+        onChange={(event) => {
+          setValue(event.target.value);
+          args.onChange?.(event);
+        }}
+      />
+    );
+  },
 };
 
 export const Size: Story = {
--- a/packages/design-system-react/src/components/TextField/TextField.stories.tsx
+++ b/packages/design-system-react/src/components/TextField/TextField.stories.tsx
@@ -50,7 +50,6 @@ const meta: Meta<TextFieldProps> = {
     truncate: { control: 'boolean' },
     placeholder: { control: 'text' },
     value: { control: 'text' },
-    className: { control: 'text' },
   },
 };
 
--- a/packages/design-system-react/src/components/TextField/TextField.stories.tsx
+++ b/packages/design-system-react/src/components/TextField/TextField.stories.tsx
@@ -1,6 +1,7 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
 import React, { useEffect, useState } from 'react';
 
+import { Text } from '../Text';
 import README from './README.mdx';
 import { TextField } from './TextField';
 import { TextFieldSize, TextFieldType } from './TextField.types';
@@ -105,7 +106,7 @@ export const StartAccessory: Story = {
   render: () => (
     <ControlledTextField
       placeholder="0.00"
-      startAccessory={<span>$</span>}
+      startAccessory={<Text>$</Text>}
       value=""
     />
   ),
@@ -115,7 +116,7 @@ export const EndAccessory: Story = {
   render: () => (
     <ControlledTextField
       placeholder="Amount"
-      endAccessory={<span>USD</span>}
+      endAccessory={<Text>USD</Text>}
       value=""
     />
   ),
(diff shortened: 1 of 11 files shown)


brianacnguyen (changes requested):
Can you address Bugbot comments?

brianacnguyen (dismissed):
do you think you can extract TextField types out to shared like the other components we've been migrating?

packages/design-system-shared/src/types/TextField/TextField.types.ts
```diff
@@ -2,18 +2,73 @@ import type { ReactNode } from 'react';
...
+   */
+  name?: string;
+  /**
+   * `id` of the inner `input` element.
+   */
+  id?: string;
+  /**
+   * If true, the inner input is marked as required.
+   */
+  required?: boolean;
```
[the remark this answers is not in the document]
brianacnguyen: @kirillzyusko let's wait for @georgewrmarshall on this one

packages/design-system-shared/src/types/TextField/TextField.types.ts
```diff
@@ -2,18 +2,73 @@ import type { ReactNode } from 'react';
...
 export type TextFieldPropsShared = Omit<
   InputPropsShared,
   'isStateStylesDisabled' | 'textVariant'
 > & {
+  /**
+   * The size of the text field. Controls the height.
+   *
+   * @default TextFieldSize.Md
+   */
+  size?: TextFieldSize;
```
georgewrmarshall: These props added to `TextFieldPropsShared` need to move to the React-only `TextFieldProps`. Cross-referencing against the RN `TextField` implementation:

- `size` — removed from RN as a breaking change in v0.12.0 (PR #1000); implementation hardcodes `h-12`
- `type` — HTML `input` attribute with no RN equivalent; RN consumers use `inputProps.keyboardType` / `inputProps.secureTextEntry` instead; falls through to `Box` via `...props`
- `truncate` — RN hardcodes `numberOfLines={1}`; prop is not consumed
- `name` / `id` / `required` — HTML form attributes with no RN equivalent; fall through to `Box` via `...props`
- `maxLength` — RN `TextInput` supports this natively but the RN `TextField` doesn't destructure or forward it to the inner `Input`, so it currently falls through to `Box` instead

The props that correctly belong in shared are: `isError`, `startAccessory`, `endAccessory`, and `inputElement`. Once the web-only props move to the React layer, `TextFieldSize` and `TextFieldType` const objects can also move to the React package since nothing in shared will reference them.

packages/design-system-react/src/components/TextField/TextField.stories.tsx
```diff
@@ -0,0 +1,163 @@
...
+export default meta;
+
+type Story = StoryObj<TextFieldProps>;
+
+export const Default: Story = {
+  args: {
+    placeholder: 'Sample placeholder',
+    value: '',
+  },
+  render: (args) => <ControlledTextField {...args} />,
```
georgewrmarshall: non-blocking: The Default story renders via `<ControlledTextField>`, so Storybook's "View Code" panel shows that wrapper name instead of `<TextField>`. Since that panel is intended to be a copy-pasteable example, consider inlining `useState` directly in the `render` function so `<TextField>` is the visible root element.

Screenshot 2026-05-20 at 12 07 12 PM

georgewrmarshall (changes requested):
The implementation and MIGRATION.md are solid, but the `TextFieldPropsShared` additions need to be addressed before merging — several props added to the shared type are either HTML-specific or were deliberately removed from the RN platform. See inline comments for the full breakdown.

One additional item not capturable as an inline comment: the focused border color differs between platforms. The React implementation uses `border-primary-default` (blue) while the RN `TextField` uses `border-default` (gray) — worth aligning

Screenshot 2026-05-20 at 12 03 45 PM
Screenshot 2026-05-20 at 12 20 03 PM

georgewrmarshall (approved):
LGTM! The two blocking issues from round 1 are properly resolved: web-only props (`TextFieldSize`, `TextFieldType`, `size`, `type`, `truncate`, `name`, `id`, `required`, `maxLength`) correctly moved out of shared into the React package, and the `TextFieldInputProps` `Omit` list now covers all props that would silently collide with top-level ones. The Default story inline state fix and `className` control removal are also done. Nice work. Approved!

PR #1210, opened 2026-06-03
refactor: Added default padding and added isInteractive to SectionHeader

brianacnguyen:
## **Description**

This PR updates the React Native `SectionHeader` component to match design specs and support tappable section headers.

**Why:** Section headers needed consistent default spacing, and some headers (e.g. "See all") should be pressable with clear visual feedback.

**What changed:**

- **Default padding** — Applies `px-4 pt-3 pb-2` to the outer wrapper by default (still overridable via `twClassName`).
- **`isInteractive` prop** — Discriminated union type:
  - `false`/omitted: static `BoxRow` with `ViewProps` (unchanged behavior).
  - `true`: wraps the header in `Pressable` with `PressableProps`; opacity reduces to `0.7` while pressed.
- **Default end icon** — When `isInteractive` is `true`, renders `IconName.ArrowRight` unless `endIconName`, `endIconProps.name`, or `endAccessory` is provided.
- **End icon color** — End icons use `IconColor.IconAlternative`.

Also includes README updates, an `Interactive` Storybook story, unit tests, and a jest branch coverage override for the untestable pressed-state branch (same pattern as `ActionListItem`).

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-772

## **Manual testing steps**

1. Run React Native Storybook: `yarn storybook:ios` or `yarn storybook:android`
2. Open **Components/SectionHeader**
3. **Default** — Confirm default padding (`px-4 pt-3 pb-2`) on static headers
4. **Interactive** — Confirm:
   - Arrow-right end icon appears by default
   - Row is tappable (`onPress` fires)
   - Opacity reduces to ~0.7 while pressed
5. Optionally verify overrides: custom `endIconName`, `endAccessory`, and `twClassName` still work when `isInteractive` is `true`

### **After**

https://github.com/user-attachments/assets/e3ef79ba-95ff-40eb-b148-ec77b8c84775

---

> [!NOTE]
> **Medium Risk**
> Default padding and optional Pressable behavior change layout and touch targets for every SectionHeader consumer; removing icon name resolution via iconProps is a subtle breaking API change.
> 
> **Overview**
> **React Native `SectionHeader`** now applies default outer padding (`px-4 pt-3 pb-2`) and supports **`isInteractive`**: when `true`, the row is a `Pressable` with `PressableProps`, default `accessibilityRole` of `button`, `opacity-70` while pressed (skipped when `disabled`), and a default **`ArrowRight`** end icon unless `endIconName` or `endAccessory` is set. Non-interactive usage stays on a static outer `BoxRow` with `ViewProps`.
> 
> **API / styling tweaks:** `SectionHeaderProps` is a discriminated union on `isInteractive`; `startIconProps` / `endIconProps` no longer accept `name` (icons come only from `startIconName` / `endIconName`). End icons use **`IconColor.IconAlternative`**. Layout is refactored so the title sits in an inner row and start/end accessories sit on the outer wrapper (or inside the pressable inner row when interactive).
> 
> Docs, Storybook (`Interactive` story), expanded unit tests, and a **Jest branch coverage** exception for `SectionHeader.tsx` (same pressed-state limitation as `ActionListItem`) are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 879e660620d4d7e7528d1e2cdb71d4d1683f8743. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/design-system-react-native/src/components/SectionHeader/SectionHeader.stories.tsx
```diff
@@ -32,13 +34,6 @@ const meta: Meta<typeof SectionHeader> = {
         'Optional end icon; when set, renders instead of endAccessory',
     },
   },
-  decorators: [
-    (Story) => (
-      <Box twClassName="w-full bg-background-default p-4">
-        <Story />
-      </Box>
-    ),
-  ],
```
georgewrmarshall, line 41: Nice. I think we should actually remove all of these types of wrappers from our stories regardless of how they sit within the page...

packages/design-system-react-native/src/components/SectionHeader/SectionHeader.stories.tsx
```diff
@@ -98,3 +93,9 @@ export const TitleAccessory: Story = {
     />
   ),
 };
+
+export const Interactive: Story = {
+  render: () => (
+    <SectionHeader title="Assets" isInteractive onPress={noopPress} />
+  ),
```
georgewrmarshall, line 111: non-blocking: Could we add the new interactive props to Storybook `argTypes`? `isInteractive` is now a public SectionHeader prop, but the controls only expose title/start/end icon options, so consumers browsing Storybook cannot toggle the new behavior from the default story.

packages/design-system-react-native/src/components/SectionHeader/SectionHeader.tsx
```diff
@@ -41,58 +48,98 @@ export const SectionHeader: React.FC<SectionHeaderProps> = ({
...
+    </BoxRow>
+  ) : null;
+
+  if (isInteractive) {
+    const isDisabled = Boolean(
+      (wrapperRest as Omit<PressableProps, 'children'>).disabled,
+    );
+
+    return (
+      <Pressable
```
georgewrmarshall, line 108: non-blocking: Since this branch renders an interactive `Pressable`, could we default `accessibilityRole` to `button` when one is not provided? Other interactive primitives in this package, like `ButtonBase`, default to `accessibilityRole="button"`, and this keeps the new full-row tap target announced as interactive by default. I'm not fully up to date with RN accessibility roles so there may be one more appropriate?

packages/design-system-react-native/src/components/SectionHeader/SectionHeader.tsx
```diff
@@ -41,58 +48,98 @@ export const SectionHeader: React.FC<SectionHeaderProps> = ({
...
+        ...titleProps,
+      }}
     >
-      {title ? (
+      {title}
+    </BoxRow>
+  ) : null;
+
+  if (isInteractive) {
+    const isDisabled = Boolean(
```
georgewrmarshall: non-blocking: Could we avoid the cast here by narrowing/destructuring the interactive props before reading `disabled`? The runtime behavior looks correct because native `disabled` is still forwarded to `Pressable`, but `(wrapperRest as Omit<PressableProps, 'children'>).disabled` is a bit hard to follow and works around the prop union rather than letting TypeScript narrow it naturally.

packages/design-system-react-native/src/components/SectionHeader/SectionHeader.types.ts
```diff
@@ -35,4 +31,17 @@ export type SectionHeaderProps = SectionHeaderPropsShared & {
...
-} & Omit<ViewProps, 'children'>;
+};
+
+/**
+ * SectionHeader component props (React Native).
+ * Extends {@link SectionHeaderPropsShared} with platform `Text` / `Icon` passthroughs, `twClassName`, and wrapper props.
+ * When `isInteractive` is `true`, the outer wrapper is a `Pressable`; otherwise it is a {@link BoxRow} with `ViewProps`.
+ */
+export type SectionHeaderProps =
+  | (SectionHeaderPropsBase & {
```
georgewrmarshall, line 42: non-blocking: Could we omit `name` from `startIconProps` and `endIconProps` and require icon names to come from `startIconName` / `endIconName` instead? Supporting both `endIconName` and `endIconProps.name` creates two ways to set the same value and requires the `finalStartIconName` / `finalEndIconName` resolution logic. Omitting `name` from the prop bags would make the API clearer and simplify the component.

packages/design-system-react-native/src/components/SectionHeader/SectionHeader.tsx
```diff
@@ -41,58 +48,98 @@ export const SectionHeader: React.FC<SectionHeaderProps> = ({
   endAccessory,
   endIconName,
   endIconProps,
+  isInteractive,
   twClassName = '',
-  ...rest
+  style,
+  ...wrapperRest
```
georgewrmarshall: nit: Could we keep this as `...props` instead of `...wrapperRest`? The wrapper target already branches below, and the shorter name matches the surrounding component style.

georgewrmarshall (dismissed):
Nice work adding the interactive SectionHeader behavior. The implementation looks good overall, Bugbot's disabled pressed-state concern appears addressed in the latest commits, and the focused SectionHeader test suite passes locally. I left a few non-blocking comments around Storybook controls, accessibility defaults, prop typing, and docs consistency.

brianacnguyen (commented):
Let me update

packages/design-system-react-native/src/components/SectionHeader/SectionHeader.stories.tsx
```diff
@@ -8,6 +8,8 @@ import { Icon } from '../Icon';
 import { SectionHeader } from './SectionHeader';
 import type { SectionHeaderProps } from './SectionHeader.types';
 
+const noopPress = () => undefined;
```
georgewrmarshall, line 11: non-blocking: this would be a good use of actions but we can do a bulk update not critical

georgewrmarshall (approved):
Approved. I left non-blocking follow-up comments.

packages/design-system-react-native/src/components/SectionHeader/SectionHeader.test.tsx
```diff
@@ -23,6 +27,54 @@ describe('SectionHeader', () => {
     jest.clearAllMocks();
   });
 
+  const flattenStyles = (
```
georgewrmarshall, line 30: non-blocking: I think this is okay to approve and merge, but could we avoid leaning further into `flattenStyles`, `findPressableStyleFn`, and `getPressableStyleFn` long term? PR #1184 moved us away from `react-test-renderer` tree traversal because it is related to RNTL implementation-detail anti-patterns: RNTL cautions against navigating the `ReactTestInstance` tree because it makes tests fragile, and its type/props queries are explicitly marked `UNSAFE` for the same reason. I have not found a reliable way to test `Pressable` pressed-state visual logic through RNTL without reaching into renderer internals, so we may want to remove these helpers and accept/document the reduced branch coverage in the PR that upgrades RNTL. References: https://callstack.github.io/react-native-testing-library/docs/advanced/testing-env#tree-navigation and https://callstack.github.io/react-native-testing-library/docs/api/queries#legacy-unit-testing-helpers

brianacnguyen: Sure. Should we do a change to remove these tips and add a rule to the testing rules?

PR #1253, opened 2026-06-21
fix(design-system-react-native): mark BottomSheetDialog pan callbacks as worklets

xavier-brochard:
## Problem

In the MetaMask mobile app, **all bottom sheets that use MDS `BottomSheetDialog` have a broken slow-drag gesture**: the sheet does not follow the finger during a slow drag. Quick swipes still dismiss or snap back, but the interactive drag-to-position behavior is broken.

## Root cause

`BottomSheetDialog` registers `Gesture.Pan()` handlers (`onStart`, `onUpdate`, `onEnd`) to track finger position during a drag. This package ships a pre-built `dist` compiled by ts-bridge (tsc), which emits the gesture chain as a namespaced call:

```js
react_native_gesture_handler_1.Gesture.Pan().onUpdate(...)
```

The consumer's Reanimated/Worklets Babel plugin **does** run over `dist` (which is why `useAnimatedStyle` open/close animations work), but its gesture auto-detection does not recognize that compiled namespaced form. Without explicit worklet marking, the pan callbacks run on the JS thread instead of the UI thread — so `onUpdate` lags behind the finger during slow drags.

## Fix

Add explicit `'worklet'` directives to the three `Gesture.Pan()` callbacks in `BottomSheetDialog.tsx`. These survive compilation into `dist` and tell the consumer's Babel plugin to workletize the handlers regardless of auto-detection.

## Test plan

- [x] `yarn workspace @metamask/design-system-react-native run test:verbose -- BottomSheetDialog --collectCoverage=false`
- [ ] Verify slow drag on `BottomSheetDialog` tracks the finger in the MetaMask mobile app
- [ ] Verify slow drag in the React Native Storybook app
- [ ] Verify dismiss swipe and snap-back behavior still work as expected

### DIFF
--- a/packages/design-system-react-native/CHANGELOG.md
+++ b/packages/design-system-react-native/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed `BottomSheetDialog` slow-drag not tracking the finger by marking `Gesture.Pan()` callbacks as Reanimated worklets in the published `dist` bundle
+
 ## [0.30.0]
 
 ### Added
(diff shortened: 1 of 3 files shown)


georgewrmarshall (approved):

georgewrmarshall:
@metamaskbot publish-preview

PR #1270, opened 2026-06-24
fix(BannerAlert): remove left border and padding override

andrewjcohen:
## Summary
- Remove the severity-colored left border (`border-l-4`) from React Native `BannerAlert` to match Figma
- Remove the `paddingLeft={2}` override so `BannerAlert` inherits `BannerBase` default padding
- Clean up unused border color constants and update unit tests

## Test plan
- [x] `npx jest BannerAlert.test.tsx --coverage=false` (6 tests passing)
- [x] Verify BannerAlert severity variants in React Native Storybook
- [x] Confirm visual alignment with [Figma BannerAlert](https://www.figma.com/design/0GipXMImYPyEmNUvnyI5v8/refactor--BannerBase---BannerAlert?node-id=16610-22779)

## Loom
https://www.loom.com/share/906b343a2e65414d9cc2f4881722bfd1

## Before
image

## After
image

### DIFF
--- a/packages/design-system-react-native/src/components/BannerAlert/BannerAlert.constants.ts
+++ b/packages/design-system-react-native/src/components/BannerAlert/BannerAlert.constants.ts
@@ -1,7 +1,6 @@
 import {
   BannerAlertSeverity,
   BoxBackgroundColor,
-  BoxBorderColor,
   IconColor,
   IconName,
 } from '@metamask/design-system-shared';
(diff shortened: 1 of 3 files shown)


amandaye0h:
Right now, it looks like we are missing a `neutral` variant in code. This is a need to have so our design/code match.

Variants are defined in `packages/design-system-shared/src/types/BannerAlert/BannerAlert.types.ts`. More generally, we keep shared props between React and React Native so they share the same logic. You can read more about this decision [here](https://github.com/MetaMask/decisions/blob/main/decisions/design-system/0004-centralized-types-architecture.md).

Screenshot 2026-06-25 at 3 58 29 PM

I'll approve this PR for now so it's unblocked, but strongly recommend these additions in a follow up PR:

1. Adding the `neutral` variant in `BannerAlert.types.ts`
2. Updating `BannerAlert` stories to document the `neutral` variation

amandaye0h (approved):
Approving so we get the polish updates in - but will need a follow up PR to address the missing `neutral` variant

georgewrmarshall (approved):
Nice sized style polish PR. Approving, but two follow-ups needed:

1. The web `BannerAlert` (`packages/design-system-react`) still applies `border-l-4` and `paddingLeft={2}` — it needs the same treatment to stay in platform alignment. For a change this size it's generally better to do both platforms in one PR.
2. Worth a follow-up for the neutral variant as well.

andrewjcohen:
PR for web banneralert changes https://github.com/MetaMask/metamask-design-system/pull/1283

PR #1372, opened 2026-07-07
feat: [DSRN] Added Slider component

brianacnguyen:
## **Description**

Adds the React Native `Slider` component to MMDS, migrated from the Perps slider pattern in metamask-mobile. This gives consumers a controlled range input with pan/tap gestures, optional range dots and labels, and hooks for non-linear scales (e.g. mUSD-style amount mapping).

**Why:** Perps and other mobile flows need a shared, design-system slider instead of maintaining a one-off implementation.

**What changed:**
- **`@metamask/design-system-shared`** — `SliderPropsShared` (ADR-0004) with cross-platform API: range config, `showRangeLabels` / `showRangeDots`, `onGrip` / `onTick`, and custom scale hooks (`mapValueToTrackPercent`, `mapTrackPercentToValue`, `formatStepLabel`, `stepToValue`)
- **`@metamask/design-system-react-native`** — Full RN implementation:
  - `Slider.tsx` — controlled UI with `Box` / `Text`, accessibility increment/decrement
  - `useSliderGesture.ts` — Reanimated + RNGH pan/tap handling on a track-percent coordinate system
  - `Slider.utilities.ts` — pure mapping/layout helpers (domain value ↔ track % ↔ pixels)
  - 28 unit tests, Storybook stories, README
  - Per-file Jest coverage threshold for `useSliderGesture.ts` (gesture worklets not unit-testable)
- **Layout fixes** — range dots render inside the track; labels center on snap points; 8px gap between thumb and labels; stacked story spacing to avoid overlap

**Known limitation:** Accessibility increment/decrement uses linear `step` on the domain value. Non-linear consumers may need custom screen-reader handling.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-689

## **Manual testing steps**

1. Run React Native Storybook: `yarn storybook:ios` (or `:android`)
2. Navigate to **Components/Slider**
3. **Default** — drag thumb, tap track; confirm value updates smoothly
4. **ShowRangeLabels** / **ShowRangeDots** — confirm dots sit on the track and labels align under snap points with ~8px gap below the thumb
5. **Step** — confirm stepped snapping (step 1 vs step 5)
6. **IsDisabled** — confirm interaction is blocked
7. **StepToValue** — confirm mUSD-style labels (`$100`, `$1,000`, `$10,000`) and non-linear thumb position when dragging/tapping
8. Run unit tests: `yarn workspace @metamask/design-system-react-native run test src/components/Slider/Slider.test.tsx`

### **Before**

N/A — new component

### **After**

https://github.com/user-attachments/assets/290a9a0f-4096-41e7-9b7f-7f61eb2579e9

---

> [!NOTE]
> **Low Risk**
> New design-system UI with no auth or data-path changes; gesture drag behavior is deferred to Storybook/device per existing BottomSheet pattern.
> 
> **Overview**
> Introduces a **controlled React Native `Slider`** for the design system (migrated from the Perps pattern), exported from `@metamask/design-system-react-native` with cross-platform **`SliderPropsShared`** in `@metamask/design-system-shared`.
> 
> The implementation supports **pan/tap** interaction via Reanimated + RNGH (`useSliderGesture`), optional **range dots/labels**, **`onGrip` / `onTick`**, and **custom scale hooks** for non-linear ranges (e.g. mUSD-style `map*` / `stepToValue`). The root uses **`accessibilityRole="adjustable"`** with increment/decrement; a11y stepping stays linear on domain values when scales are non-linear.
> 
> Also adds **unit tests**, **Storybook** stories, **README**, and **Jest** tweaks: per-file coverage thresholds for `useSliderGesture.ts`, and ignoring `*.types.ts` in coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e98c3a6f9ace23f2205fbb5b6e326689d26ef0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xavier-brochard:
Thanks for adding this! We'll definitely want to use it in the QuickBuy in place of our custom one.

Just a gap I'm seeing: could we have an option to disable the snapping while still showing the 25% / 50% / ... markers? Otherwise it would prevent user to be as precise as they need. Markers would still show as pure visual clues.

Also suggestion: what about making the slider's dot animate / get bigger when it's gripped? Nice visual clue that it's effectively gripped

brianacnguyen:
> Thanks for adding this! We'll definitely want to use it in the QuickBuy in place of our custom one.
> 
> Just a gap I'm seeing: could we have an option to disable the snapping while still showing the 25% / 50% / ... markers? Otherwise it would prevent user to be as precise as they need. Markers would still show as pure visual clues.
> 
> Also suggestion: what about making the slider's dot animate / get bigger when it's gripped? Nice visual clue that it's effectively gripped

There is currently no "snapping" built in for the slider. the user can press on the labels to "snap" to the point on the slider, but the slider itself moves based on the step defined

packages/design-system-react-native/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,220 @@
...
+      onDragEnd,
+      onValueChange,
+      step,
+      value,
+    ],
+  );
+
+  return (
+    <Box
+      {...props}
```
georgewrmarshall, line 125: The `pb-[30px]` arbitrary value violates the no-arbitrary-values rule in `.cursor/rules/styling.md` ("NEVER use arbitrary values unless critical"). The value comes from geometry that's already in constants: `RANGE_LABEL_TOP = SLIDER_VERTICAL_PADDING + THUMB_BOTTOM_OFFSET + THUMB_TO_LABEL_GAP = 35`. The bottom padding only needs to cover `RANGE_LABEL_TOP` plus approximate text height (~20px), so `pb-[30px]` may even be slightly short. Consider deriving it as a constant in `Slider.constants.ts` (e.g. `SLIDER_LABEL_PADDING_BOTTOM = RANGE_LABEL_TOP`) and referencing it via the `style` prop rather than `twClassName`.

Separately, the `mx-4` on the same line bakes a 16px horizontal margin into every consumer's layout. The geometry reason is clear — the thumb overhangs 16px at min/max via `THUMB_LEFT_OFFSET = -16`, and `hitSlop` needs physical space — but a caller who already has container padding will end up with 32px of edge spacing. The README mentions wrapping in `GestureHandlerRootView` but is silent on the implicit margin requirement. At minimum, document this constraint (and that it can be overridden with the `style` prop); ideally expose a `trackInset` prop so the geometry and the layout are controlled together.

packages/design-system-shared/src/types/Slider/Slider.types.ts
```diff
@@ -0,0 +1,115 @@
...
+  showRangeLabels?: boolean;
+
+  /**
+   * When true, renders dots on the track at each rangeLabelSteps position.
+   *
+   * @default false
+   */
+  showRangeDots?: boolean;
+
+  /**
```
georgewrmarshall, line 70: non-blocking: The JSDoc says "Fired when the user begins or ends dragging the thumb" but doesn't make clear it fires **twice per drag** — once on `Pan.onStart` and once inside `commitAtPosition` on `Pan.onEnd`. Haptic consumers will therefore call their feedback API twice per gesture, which is the intended mobile behavior (matching `ImpactMoment.SliderGrip` on begin and end). Worth being explicit: "Fires once when pan begins and once when pan ends; not fired on tap or range-label press."

packages/design-system-react-native/src/components/Slider/Slider.types.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Wire `handleLayout`, animated styles, and `gesture` into `Slider.tsx`;
+ * `handlePressStep` is bound to range-label `Pressable` targets.
+ */
+export type UseSliderGestureResult = {
+  /**
+   * Root `onLayout` handler. Stores track width and re-syncs thumb position
+   * from the current `value` prop.
+   */
+  handleLayout: (event: { nativeEvent: { layout: { width: number } } }) => void;
+
```
georgewrmarshall, line 134: suggestion: Using inline dynamic import types (`ReturnType<typeof import('react-native-reanimated').useAnimatedStyle>`) is unconventional and harder to navigate in IDE tooling. Adding `import type { useAnimatedStyle } from 'react-native-reanimated'` and `import type { Gesture } from 'react-native-gesture-handler'` at the top and using `ReturnType<typeof useAnimatedStyle>` is cleaner. The fact that `Slider.types.ts` would then import from Reanimated and RNGH is not a problem — both are already peer deps of the package.

packages/design-system-react-native/jest.config.js
```diff
@@ -67,6 +76,7 @@ module.exports = merge(baseConfig, {
     '\\.constants\\.ts$', // ignore all .constants.ts files
     '\\.dev\\.ts$', // ignore all .dev.ts files
     '\\.figma\\.tsx', // Figma code connect files
+    '\\.types\\.ts$', // type-only modules
     'Icon.assets.ts', // Exclude Icon.assets file
```
georgewrmarshall, line 80: non-blocking: `'\.types\.ts$'` is added to the global ignore list, which will exclude every `*.types.ts` file across the entire package — not just Slider. The intent matches project convention (`.types.ts` should be type-only, no runtime code), but it's worth a comment on the line making this explicit, similar to how `'\.constants\.ts$'` is documented. If any existing `.types.ts` file happens to export a const object alongside types (e.g. before a migration is completed), it would silently become uncovered.

packages/design-system-react-native/src/components/Slider/Slider.utilities.ts
```diff
@@ -0,0 +1,266 @@
...
+): number {
+  return (step / 100) * (maximumValue - minimumValue) + minimumValue;
+}
+
+/**
+ * Default `formatStepLabel`: display track percent as a percent string.
+ *
+ * @param step - Track-percent step from `rangeLabelSteps`.
+ * @returns Formatted label string.
+ */
```
georgewrmarshall, line 225: non-blocking: `getDotLeftPercent` hard-codes `'2%'` and `'98%'` for the edge steps. The `2%` offset is meant to keep the dot visually on the track when the thumb is at position 0, but the magic values are unexplained. A comment connecting this to `THUMB_LEFT_OFFSET` / `THUMB_SIZE` geometry (or a named constant like `DOT_EDGE_INSET_PERCENT`) would help future readers understand why the edge steps get a different value than `step%`.

georgewrmarshall (commented):
Approved! Looking good. A few issues worth addressing, mostly around a styling-rule violation and some documentation gaps in the public API.

georgewrmarshall (dismissed):

packages/design-system-react-native/src/components/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,201 @@
...
+};
+
+export const Step: Story = {
+  render: () => (
+    <Box gap={12}>
+      <ControlledSlider value={50} step={1} showRangeLabels showRangeDots />
+      <ControlledSlider value={50} step={5} showRangeLabels showRangeDots />
+    </Box>
+  ),
+};
```
georgewrmarshall, line 126: non-blocking: I can't tell what the difference is that we are showing between these sliders they look exactly the same to me?

Image

georgewrmarshall (approved):
LGTM! One non-blocking comment that I missed in first review

packages/design-system-react-native/src/components/Slider/Slider.stories.tsx
brianacnguyen: the difference is the step. one is more sensitive than the other

PR #1397, opened 2026-07-21
fix(dsrn): stop Slider thumb rewind on rapid taps and fast pans

brianacnguyen:
## **Description**

Rapid taps, range-label presses, or fast pans on React Native `Slider` could make the thumb briefly jump backward (and could false-fire or miss `onMark`): the UI thread already moved to the newer position, then a lagged controlled-`value` prop echo from an earlier `onValueChange` overwrote `propValue` / `translateX` (and could rewind the haptic baseline) before the newer echo caught up.

**What changed:**
- In `useSliderGesture`, each direct commit (tap / pan / label) records a monotonic **commit generation** with the emitted domain value — same sequence principle as React Native `TextInput`'s `mostRecentEventCount`, kept local so the public API stays a plain `value` number
- Incoming `value` props are reconciled against recent commits: matching emits are acknowledged without rewinding the thumb or haptic baseline (including **late** echoes after a newer commit was already acked); non-matching values are treated as external updates and applied immediately
- Stale echoes are not written into `propValue`, and `hasInflightCommits` guards `useAnimatedReaction` while newer local commits are unacked
- Hook tests cover stale in-flight echoes, external updates while commits are pending, latest-ack behavior, and late echoes after the parent already caught up
- No public `Slider` API changes — drop-in behavioral fix for `@metamask/design-system-react-native`

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-933

## **Manual testing steps**

1. From repo root, run `yarn storybook:ios` (or `yarn storybook:android`) and open **Components → Slider → ShowRangeLabels**
2. Rapidly tap different points on the track in quick succession. Confirm the thumb lands on the last tap without snapping backward
3. Rapidly tap range labels (`0%` → `100%` → `25%` → `75%` …). Same check — no backward flicker
4. Drag the thumb **quickly** across the track and release. Confirm no backward flicker during the scrub or on release
5. Change `value` from outside (Storybook controls or a programmatic reset). Confirm the thumb still syncs to external updates (including while a local gesture echo may still be in flight)
6. Optional unit tests: `yarn workspace @metamask/design-system-react-native exec NODE_OPTIONS=--experimental-vm-modules jest src/components/Slider --watchman=false`

### **Before**

https://github.com/user-attachments/assets/7fcb81df-1844-41eb-9757-693b7ae566d5

https://github.com/user-attachments/assets/7fda9351-914d-4393-b8e1-ef0d379d4f66

### **After**

https://github.com/user-attachments/assets/97c4bb95-8b69-48ea-a95f-dcb97c489051

---

> [!NOTE]
> **Medium Risk**
> Touches controlled-value sync, Reanimated reactions, and haptic threshold state in a shared design-system hook; behavior is heavily tested but regressions could affect any controlled Slider usage.
> 
> **Overview**
> Fixes controlled **Slider** thumb flicker and incorrect **`onMark`** firing when parent `value` echoes arrive out of order after fast taps, label presses, or pans.
> 
> **`useSliderGesture`** now tracks each local emit with a monotonic **commit generation** (similar in spirit to TextInput event sequencing) and reconciles incoming `value` props against a bounded recent-commit history. Stale self-echoes are acknowledged without updating **`propValue`**, the haptic baseline, or **`translateX`**; values that do not match recent emits still apply immediately as external updates. **`hasInflightCommits`** blocks **`useAnimatedReaction`** from moving the thumb while newer commits are unacked, and **`handleLayout`** remaps the current thumb on resize instead of re-syncing from a lagged raw `value` prop. Range-label taps route through **`emitValueChange`** so commits are recorded consistently.
> 
> Adds hook tests for stale in-flight echoes, external updates during pending commits, latest-ack behavior, late echoes after the parent caught up, and layout remapping. No public **Slider** API changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f94ef721f88bb6b59c7405a76067b364fbcdf98c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


georgewrmarshall:
Great before/after videos, they show the bug clearly and make it easy to verify the fix visually. That said, I think the solution approach needs a rethink.

I understand the bug: rapid taps cause stale `value`-prop echoes to briefly snap the thumb backward because the JS-thread round trip (gesture → `onValueChange` → parent state → re-render → `useAnimatedReaction`) completes after a newer commit has already advanced the UI thread.

The fix direction is right. The problem is the mechanism chosen to suppress stale echoes.

**`SELF_ECHO_GRACE_MS` — wrong tool for this problem**

Time-based suppression is a heuristic that fails at both ends:

- **Slow devices:** A JS-thread round trip on low-end Android under load can exceed 120ms. The stale echo arrives after the window, the thumb still snaps back, and the bug is silently unreproducible in dev.
- **External updates are permanently dropped:** If a programmatic `value` change arrives within 120ms of a tap, `useAnimatedReaction` skips the `translateX` update (line 171) and `previousTrackPercentRef` is not updated (line 137). No subsequent prop change is guaranteed to come, that update is silently lost.

The constant's own JSDoc frames this clearly: "Keep this comfortably above a slow JS-thread round trip while staying under normal human tap cadence." That's not a guarantee of correctness, it's a bet on timing.

React Native solved this exact problem, same JS/UI thread async boundary, same controlled component echo, not with a timer, but with a monotonically increasing sequence number. Worth looking at how [`TextInput` handles `mostRecentEventCount`](https://github.com/react/react-native/blob/v0.81.5/packages/react-native/Libraries/Components/TextInput/TextInput.js#L397) and exploring whether the same principle can be adapted here

packages/design-system-react-native/src/components/Slider/useSliderGesture.ts
```diff
@@ -86,6 +116,77 @@ export function useSliderGesture(
...
+      }
+
+      if (match) {
+        // Late echo of an already-acked commit — do not write propValue or the
+        // reaction will snap the thumb backward after a fast pan.
+        if (match.generation <= ackedGenerationRef.current) {
+          return { syncPropValue: false, syncBaseline: false };
+        }
+
+        if (incoming === lastEmittedValueRef.current) {
```
georgewrmarshall, line 174: non-blocking: Would a self-review comment help explain why `lastEmittedValueRef` is compared here rather than just using `match.generation` in both branches? Is that when the echo matches the latest emit you want to jump `ackedGenerationRef` all the way to `lastEmittedGenerationRef` to clear inflight faster (skipping any intermediate echoes still outstanding), but a brief comment confirming that would make this easier to reason about at a glance.

packages/design-system-react-native/src/components/Slider/useSliderGesture.ts
```diff
@@ -158,9 +281,10 @@ export function useSliderGesture(
 
   const emitValueChange = useCallback(
     (nextValue: number) => {
+      recordDirectCommit(nextValue);
```
georgewrmarshall, line 312: non-blocking: `emitValueChange` is called on every `onUpdate` during a pan drag, so each intermediate position now creates a commit entry. That looks intentional, but a short comment noting it would help future readers understand why the buffer can accumulate many entries after a fast drag, and why 256 was chosen as the limit rather than a smaller number.

georgewrmarshall (dismissed):
LGTM! Tested this on device, the thumb snap back and haptic baseline rewind are both fixed. Nice work 👍 

When I suggested looking at React Native TextInput's `mostRecentEventCount` I didn't anticipate we'd end up this deep 😅

Looking more closely the key difference is that TextInput's counter is bidirectional with native, JS echoes the count back so native validates and drops stale events itself. Without that native round trip, the full suppression logic has to live in JS, which is why the implementation is more involved.

Approving with a couple of non-blocking questions on the trickiest parts of `reconcilePropValue`.  A reminder that self review comments explaining complex areas of the code make reviews much easier!

packages/design-system-react-native/src/components/Slider/useSliderGesture.ts
brianacnguyen: Good catch. If the incoming value matches the latest value we sent via onValueChange, we treat the parent as fully caught up and clear any older pending echoes in one go (React often batches those during a fast pan, so they may never show up individually). If it matches an older value instead, we only mark that older update as handled and keep waiting for the newer ones

georgewrmarshall (dismissed):

packages/design-system-react-native/src/components/Slider/useSliderGesture.ts
```diff
@@ -394,6 +546,7 @@ export function useSliderGesture(
       }
 
       const newValue = getMarkValue(mark, minimumValue, maximumValue);
+      recordDirectCommit(newValue);
```
georgewrmarshall: non-blocking: `handlePressStep` calls `recordDirectCommit(newValue)` + `onValueChange(newValue)` directly, which is exactly what `emitValueChange` does. Routing through `emitValueChange` here would remove the duplication and ensure any future changes to the emit flow (e.g. logging, throttling) apply automatically to both paths.

georgewrmarshall (dismissed):
Solid fix. The commit generation + ack approach mirrors React Native TextInput's `mostRecentEventCount` pattern and is a reliable solution to the controlled-value echo problem. The tests cover all the meaningful edge cases described in the PR. A couple of non-blocking comments.

georgewrmarshall (approved):

PR #1403, opened 2026-07-22
ci: reduce workflow overhead with low-risk cleanups (pt 1)

georgewrmarshall:
## **Description**

A set of low-risk, no-behavior-change cleanups to the GitHub Actions workflows. These were identified as part of a broader workflow optimization audit; a follow-up PR (pt 2) will tackle the more structural changes (artifact sharing, Storybook workflow consolidation).

Changes:

- **Remove single-value `node-version` matrices** from `lint-build-test.yml` and `chromatic.yml`. A matrix with one entry still pays the scheduling/fanout overhead with zero parallelism benefit. `cache-node-modules` in the `prepare` job was updated from the now-removed `matrix.node-version == '24.x'` condition to a plain `true`.
- **Extract `yarn test:scripts` into its own job** (`test-scripts`) so it runs once instead of once per workspace package in the `test` matrix.
- **Remove `action-checkout-and-setup` from the label-check workflow.** The `github-script` action reads labels directly from `context.payload.pull_request.labels`, which is always present in `pull_request` events — no local checkout is needed.
- **Run `analyse-code` in parallel with `check-workflows`** by removing the `needs: check-workflows` dependency. The security scanner has no dependency on workflow file validation and was being blocked unnecessarily.

## **Expected CI impact**

This PR delivers real efficiency gains, but **don't expect a dramatic drop in overall Main workflow wall-clock time**. Comparing recent runs on this branch vs main, total time was within normal variance (~217s vs ~221s).

### What we gain

- **Less duplicate work:** `yarn test:scripts` previously ran 6× (once per package matrix job); it now runs once in a dedicated `test-scripts` job (~40–50s of redundant runner time removed per PR).
- **Faster package test jobs:** Removing `test:scripts` from each matrix leg trimmed the slowest jobs by ~9–18s (e.g. react-native ~104s → ~95s, react ~93s → ~75s).
- **Parallel security scan:** `analyse-code` no longer waits for `check-workflows` (~7s earlier start on a ~3 min job).
- **Minor trims:** Single-entry matrix removal and dropping checkout from the label-check workflow reduce scheduling/checkout overhead (seconds, not minutes).

### Why total time barely changes (for now)

The Main workflow critical path is still dominated by jobs this PR doesn't shorten:

- Code scanner (~3 min)
- Slowest package test (~1.5 min + prepare), still running a full `yarn build` per matrix job
- Chromatic and Storybook accessibility tests

Pt 1 removes wasted repetition and starts the scanner slightly sooner, but **does not** address the structural cost drivers (duplicate builds across jobs, lack of artifact sharing, no changed-path package filtering).

### Pt 2 (follow-up)

The next workflow cleanup PR will tackle those structural changes — artifact sharing between jobs, Storybook workflow consolidation, and related optimizations — where the meaningful wall-clock improvements should show up.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Verify CI passes on this PR — all jobs in the `Main` workflow should complete successfully.
2. Confirm the `test-scripts` job appears as a distinct job in the workflow run (not grouped inside the `test` matrix).
3. Confirm `ensure-blocking-pr-labels-absent` still correctly fails when the `DO-NOT-MERGE` label is applied.
4. Confirm `analyse-code` and `check-workflows` now start simultaneously in the workflow graph.

- [ ] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs)
- [ ] I've completed the PR template to the best of my ability
- [ ] I've included tests if applicable
- [ ] I've documented my code using [JSDoc](https://jsdoc.app/) format if applicable

---

> [!NOTE]
> **Low Risk**
> Changes are limited to GitHub Actions orchestration with no application or security logic changes; the main risk is mis-wiring jobs so a check is skipped or duplicated, which the PR author describes as behavior-preserving.
> 
> **Overview**
> Trims GitHub Actions scheduling overhead without changing what gets validated.
> 
> **`lint-build-test.yml` and `chromatic.yml`:** Drops single-entry `node-version` matrices from jobs that never parallelized. The `prepare` job now always passes `cache-node-modules: true` instead of tying caching to the removed matrix.
> 
> **Script tests:** Adds a dedicated **`test-scripts`** job that runs `yarn build` and `yarn test:scripts` once after `prepare`, and removes that step from each **`test`** matrix leg so script tests are not repeated per workspace package.
> 
> **Label gate:** **`ensure-blocking-pr-labels-absent`** no longer runs `action-checkout-and-setup`; it only uses `github-script` against `context.payload.pull_request.labels`.
> 
> **`main.yml`:** **`analyse-code`** no longer `needs: check-workflows`, so the security scanner can start alongside workflow linting instead of waiting on it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 237458f6a8913c84a0f970351c863a58fc4846cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -10,9 +10,6 @@ jobs:
   chromatic:
     name: Run Chromatic
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [24.x]
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
--- a/.github/workflows/ensure-blocking-pr-labels-absent.yml
+++ b/.github/workflows/ensure-blocking-pr-labels-absent.yml
@@ -13,10 +13,6 @@ jobs:
     permissions:
       pull-requests: read
     steps:
-      - name: Checkout and setup environment
-        uses: MetaMask/action-checkout-and-setup@v1
-        with:
-          is-high-risk-environment: false
       - name: Run command
         uses: actions/github-script@v7
         with:
--- a/.github/workflows/lint-build-test.yml
+++ b/.github/workflows/lint-build-test.yml
@@ -7,9 +7,6 @@ jobs:
   prepare:
     name: Prepare
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [24.x]
     outputs:
       child-workspace-package-names: ${{ steps.workspace-package-names.outputs.child-workspace-package-names }}
     steps:
@@ -17,7 +14,7 @@ jobs:
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: false
-          cache-node-modules: ${{ matrix.node-version == '24.x' }}
+          cache-node-modules: true
       - name: Fetch workspace package names
         id: workspace-package-names
         run: |
@@ -28,9 +25,6 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     needs: prepare
-    strategy:
-      matrix:
-        node-version: [24.x]
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
@@ -54,7 +48,6 @@ jobs:
     needs: prepare
     strategy:
       matrix:
-        node-version: [24.x]
         package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
     steps:
       - name: Checkout and setup environment
@@ -74,9 +67,6 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     needs: prepare
-    strategy:
-      matrix:
-        node-version: [24.x]
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
@@ -91,14 +81,10 @@ jobs:
             exit 1
           fi
 
-  test:
-    name: Test
+  test-scripts:
+    name: Test scripts
     runs-on: ubuntu-latest
     needs: prepare
-    strategy:
-      matrix:
-        node-version: [24.x]
-        package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
@@ -106,6 +92,27 @@ jobs:
           is-high-risk-environment: false
       - run: yarn build
       - run: yarn test:scripts
+      - name: Require clean working directory
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty at end of job"
+            exit 1
+          fi
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    needs: prepare
+    strategy:
+      matrix:
+        package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v1
+        with:
+          is-high-risk-environment: false
+      - run: yarn build
       - run: yarn workspace ${{ matrix.package-name }} run test
       - name: Require clean working directory
         shell: bash
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,6 @@ jobs:
 
   analyse-code:
     name: Code scanner
-    needs: check-workflows
     uses: ./.github/workflows/security-code-scanner.yml
     permissions:
       actions: read


.github/workflows/chromatic.yml
```diff
@@ -10,9 +10,6 @@ jobs:
   chromatic:
     name: Run Chromatic
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [24.x]
     steps:
```
georgewrmarshall, line 13: A GitHub Actions matrix with a single entry (here, just `24.x`) doesn't create any parallelism — it spins up exactly one job, the same as if there were no matrix at all. GitHub still pays the scheduling overhead of setting up a matrix context, evaluating the matrix, and creating a matrix job record, which adds a small but real delay. Since Node.js version is already managed by the `MetaMask/action-checkout-and-setup` action internally, declaring it here had no effect on which version was actually installed.

.github/workflows/ensure-blocking-pr-labels-absent.yml
```diff
@@ -13,10 +13,6 @@ jobs:
     permissions:
       pull-requests: read
     steps:
-      - name: Checkout and setup environment
-        uses: MetaMask/action-checkout-and-setup@v1
-        with:
-          is-high-risk-environment: false
       - name: Run command
```
georgewrmarshall, line 16: The `actions/github-script` action runs JavaScript directly inside the GitHub Actions runner and has full access to `context.payload` — the complete webhook payload that GitHub sent when this `pull_request` event fired. That payload already includes the PR's label list at `context.payload.pull_request.labels`, so there is nothing to fetch and no repo files to read. Checking out the repository (which clones all files, installs Node, and runs setup scripts) was pure overhead — roughly 30–60 seconds — for a job that only needs to inspect a value GitHub already provided.

.github/workflows/lint-build-test.yml
```diff
@@ -7,17 +7,14 @@ jobs:
...
-        node-version: [24.x]
     outputs:
       child-workspace-package-names: ${{ steps.workspace-package-names.outputs.child-workspace-package-names }}
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: false
-          cache-node-modules: ${{ matrix.node-version == '24.x' }}
+          cache-node-modules: true
```
georgewrmarshall, line 17: Previously this was `${{ matrix.node-version == '24.x' }}`, which evaluated to `true` on every run because `24.x` was the only matrix entry. The expression existed to allow selective caching in a hypothetical multi-version matrix (e.g. cache only the primary version, not each experimental one), but with a single-entry matrix it was always true. Hardcoding `true` here makes the intent explicit — node modules are always cached in the prepare job — without changing any runtime behaviour.

.github/workflows/lint-build-test.yml
```diff
@@ -91,21 +81,38 @@ jobs:
             exit 1
           fi
 
+  test-scripts:
```
georgewrmarshall, line 84: `yarn test:scripts` tests repo-level scripts (not individual packages) and produces the same result regardless of which package is being tested. Previously it lived inside the `test` matrix job, so it ran once for every workspace package — if you have 6 packages, it ran 6 times in parallel on 6 separate runners, all doing identical work. Extracting it here means it runs exactly once. The `test` matrix jobs and `test-scripts` all run concurrently off the same `prepare` gate, so there is no added latency on the critical path.

.github/workflows/lint-build-test.yml
```diff
@@ -91,21 +81,38 @@ jobs:
...
-        node-version: [24.x]
         package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: false
       - run: yarn build
-      - run: yarn test:scripts
       - run: yarn workspace ${{ matrix.package-name }} run test
```
georgewrmarshall, line 116: The risk profile here is low but worth stating: `test-scripts` and the per-package `test` jobs now run concurrently without a dependency between them. If `test:scripts` were to produce side effects that individual package tests depended on, this would be a problem. In practice `yarn test:scripts` only reads files and runs assertions — it does not write build outputs — so concurrent execution is safe. If that ever changes (e.g. a script that generates fixtures), `test` would need to declare `needs: test-scripts`.

.github/workflows/main.yml
```diff
@@ -21,7 +21,6 @@ jobs:
 
   analyse-code:
```
georgewrmarshall, line 22: Previously `analyse-code` had `needs: check-workflows`, meaning the security scan couldn't start until actionlint had finished validating every workflow file. There is no logical dependency between the two — a valid workflow file is not a prerequisite for scanning source code for security issues. Removing the dependency lets both jobs start the moment the workflow run is triggered, shaving actionlint's runtime off the security scan's start time. The `all-jobs-complete` fan-in still waits for `analyse-code` to finish before the overall run is considered passing, so nothing is skipped.

brianacnguyen (approved):

PR #1410, opened 2026-07-23
feat(dsrn): add KeyValueSelect pressable row with SelectButton value

brianacnguyen:
## **Description**

Adds a React Native `KeyValueSelect` composer for pressable key/value rows that open a picker or bottom sheet, and aligns `KeyValueRow` horizontal inset with that full-width list layout.

**What changed:**
- New `@metamask/design-system-react-native` `KeyValueSelect`: outer `Pressable` (ListItem-style `bg-pressed`, `forwardRef`) wrapping `KeyValueRow`, with a non-interactive `SelectButton` value (`pointerEvents="none"`, fixed `Md` + `Secondary`)
- KeyValueRow-style value props (`value`, `valueStartAccessory`, `valueEndAccessory`, `valueTextProps`) map onto SelectButton; SelectButton-only options live on required `selectButtonProps` (`placeholder`, caret controls)
- Component-scoped `KeyValueSelectVariant` alias (instead of importing `KeyValueRowVariant`)
- `keyValueRowProps` escape hatch for inner row shell props (`testID`, `twClassName`, `style`, a11y `ViewProps`)
- Pressable is full-bleed (`w-full`); horizontal inset comes from `KeyValueRow`, with `pr-1` merged so SelectButton’s `px-3` still totals a 16px trailing content inset
- Shared types in `@metamask/design-system-shared`, plus README, Storybook stories, and unit tests

**Breaking change:**
- `KeyValueRow` now defaults to `px-4` (16px horizontal padding). Call sites that already wrap rows in `px-4` / `paddingHorizontal={4}` will double-pad unless that wrapper padding is removed. Flush rows can override with `twClassName="px-0"`.
- See [MIGRATION.md — From version 0.37.0 to 0.38.0](packages/design-system-react-native/MIGRATION.md#from-version-0370-to-0380)

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-967

## **Manual testing steps**

1. From repo root: `yarn workspace @metamask/design-system-react-native run jest 'KeyValueRow|KeyValueSelect'`
2. Start RN Storybook (`yarn storybook:ios` or `yarn storybook:android`)
3. Open **Components → KeyValueRow** — confirm default 16px horizontal inset without a padded story wrapper; check flush override if needed (`twClassName="px-0"`)
4. Open **Components → KeyValueSelect** — **Default**: full-width pressable, 16px content inset, secondary SelectButton value, press shows `bg-pressed` (no SelectButton scale/press)
5. Spot-check **Variant**, **ValueStartAccessory**, **ValueEndAccessory**, **SelectButtonProps**, **KeyValueRowProps**, **KeyEndButtonIconProps**, and **IsDisabled** (disabled row does not fire `onPress`)

### **After**

https://github.com/user-attachments/assets/4232bfaa-c86e-42f6-b575-2c3216582f40

https://github.com/user-attachments/assets/496f41c4-1a97-40d0-9e3e-1a13f5cc339e

---

> [!NOTE]
> **Medium Risk**
> The KeyValueRow default padding change is a breaking visual/layout change for existing consumers; KeyValueSelect is new surface area but well-tested and documented.
> 
> **Overview**
> Introduces **`KeyValueSelect`**, a full-width pressable row for picker/sheet flows: outer `Pressable` with ListItem-style pressed feedback wraps **`KeyValueRow`** and a non-interactive **`SelectButton`** value (`Md` + `Secondary`). Key/value-side props mirror `KeyValueRow`; caret and placeholder live on required **`selectButtonProps`**. Shared types export **`KeyValueSelectVariant`**; **`keyValueRowProps`** forwards row shell props. Inner row merges **`pr-1`** so trailing inset stays 16px with SelectButton padding.
> 
> **Breaking:** **`KeyValueRow`** now defaults to **`px-4`** horizontal padding. Remove duplicate parent `px-4` wrappers to avoid double inset, or use **`twClassName="px-0"`** for flush rows. **0.37.0 → 0.38.0** migration notes added.
> 
> Docs, Storybook, and tests cover both components; `KeyValueRow` stories no longer pad the decorator.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26a005874b642456f90b4f40baf65f437bf55604. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.test.tsx
+++ b/packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.test.tsx
@@ -129,7 +129,23 @@ describe('KeyValueSelect', () => {
         />,
       );
 
-      expect(getByTestId('key-value-row')).toHaveStyle(tw.style('opacity-50'));
+      expect(getByTestId('key-value-row')).toHaveStyle(
+        tw.style('pl-4 pr-1', 'opacity-50'),
+      );
+    });
+
+    it('keeps SelectButton trailing inset via pr-1 on the inner KeyValueRow', () => {
+      const { getByTestId } = render(
+        <KeyValueSelect
+          keyLabel="Network"
+          selectButtonProps={{ placeholder: 'Select' }}
+          keyValueRowProps={{ testID: 'key-value-row' }}
+          onPress={noopPress}
+        />,
+      );
+
+      // KeyValueRow default px-4 + KeyValueSelect pr-1 → pl-4 pr-1
+      expect(getByTestId('key-value-row')).toHaveStyle(tw.style('pl-4 pr-1'));
     });
   });
 
@@ -218,7 +234,7 @@ describe('KeyValueSelect', () => {
   });
 
   describe('when rendering the root', () => {
-    it('applies full-width edge padding (16px leading, 4px trailing)', () => {
+    it('applies full-width pressable without horizontal padding', () => {
       const { getByTestId } = render(
         <KeyValueSelect
           keyLabel="Network"
@@ -228,12 +244,10 @@ describe('KeyValueSelect', () => {
         />,
       );
 
-      expect(getByTestId(ROOT_TEST_ID)).toHaveStyle(
-        tw.style('w-full pl-4 pr-1'),
-      );
+      expect(getByTestId(ROOT_TEST_ID)).toHaveStyle(tw.style('w-full'));
     });
 
-    it('merges twClassName with default root padding', () => {
+    it('merges twClassName with default root width', () => {
       const { getByTestId } = render(
         <KeyValueSelect
           keyLabel="Network"
@@ -245,7 +259,7 @@ describe('KeyValueSelect', () => {
       );
 
       expect(getByTestId(ROOT_TEST_ID)).toHaveStyle(
-        tw.style('w-full pl-4 pr-1', 'rounded-lg'),
+        tw.style('w-full', 'rounded-lg'),
       );
     });
 
@@ -261,7 +275,7 @@ describe('KeyValueSelect', () => {
       );
 
       expect(getByTestId(ROOT_TEST_ID)).toHaveStyle(
-        tw.style('w-full pl-4 pr-1', 'bg-pressed'),
+        tw.style('w-full', 'bg-pressed'),
       );
     });
 
@@ -277,7 +291,7 @@ describe('KeyValueSelect', () => {
       );
 
       expect(getByTestId(ROOT_TEST_ID)).toHaveStyle([
-        tw.style('w-full pl-4 pr-1'),
+        tw.style('w-full'),
         { marginTop: 8 },
       ]);
     });
--- a/packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.test.tsx
+++ b/packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.test.tsx
@@ -1,0 +1,219 @@
+import {
+  SelectButtonEndArrow,
+  SelectButtonSize,
+  SelectButtonVariant,
+} from '@metamask/design-system-shared';
+import { useTailwind } from '@metamask/design-system-twrnc-preset';
+import { fireEvent, render, renderHook } from '@testing-library/react-native';
+import React from 'react';
+import { View } from 'react-native';
+
+import { KeyValueSelect } from './KeyValueSelect';
+
+const ROOT_TEST_ID = 'key-value-select';
+const noopPress = () => undefined;
+const defaultSelectButtonProps = { placeholder: 'Select network' };
+
+describe('KeyValueSelect', () => {
+  let tw: ReturnType<typeof useTailwind>;
+
+  beforeAll(() => {
+    tw = renderHook(() => useTailwind()).result.current;
+  });
+
+  describe('when rendering content', () => {
+    it('renders keyLabel and placeholder', () => {
+      const { getByText } = render(
+        <KeyValueSelect
+          keyLabel="Network"
+          selectButtonProps={defaultSelectButtonProps}
+          onPress={noopPress}
+          testID={ROOT_TEST_ID}
+        />,
+      );
+
+      expect(getByText('Network')).toBeOnTheScreen();
+      expect(getByText('Select network')).toBeOnTheScreen();
+    });
+
+    it('renders value when set', () => {
+      const { getByText, queryByText } = render(
+        <KeyValueSelect
+          keyLabel="Network"
+          value="Ethereum Mainnet"
+          selectButtonProps={defaultSelectButtonProps}
+          onPress={noopPress}
+        />,
+      );
+
+      expect(getByText('Ethereum Mainnet')).toBeOnTheScreen();
+      expect(queryByText('Select network')).toBeNull();
+    });
+
+    it('maps valueStartAccessory to SelectButton startAccessory', () => {
+      const { getByTestId } = render(
+        <KeyValueSelect
+          keyLabel="Asset"
+          valueStartAccessory={<View testID="start-accessory" />}
+          selectButtonProps={{ placeholder: 'Select' }}
+          onPress={noopPress}
+        />,
+      );
+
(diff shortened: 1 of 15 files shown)


packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.tsx
```diff
@@ -0,0 +1,92 @@
...
+} from 'react-native';
+
+import { KeyValueRow } from '../KeyValueRow';
+import { SelectButton } from '../SelectButton';
+
+import type { KeyValueSelectProps } from './KeyValueSelect.types';
+
+const ROOT_TW_CLASS_NAME = 'w-full pl-4 pr-1';
+
+export const KeyValueSelect = ({
```
georgewrmarshall: The component creation checklist requires forwarding refs (`forwardRef`). Without it, consumers can't attach a ref to the outer Pressable for programmatic press, layout measurement, or scroll-into-view. Pattern:

```tsx
export const KeyValueSelect = forwardRef<View, KeyValueSelectProps>((
  { keyLabel, ...rest },
  ref,
) => (
  <Pressable ref={ref} ...>
    ...
  </Pressable>
));
```

packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+          value="Ethereum"
+          selectButtonProps={{ placeholder: 'Select' }}
+          onPress={noopPress}
+        />,
+      );
+
+      const selectButton = root.findByProps({ pointerEvents: 'none' });
+
+      expect(selectButton.props.size).toBe(SelectButtonSize.Md);
+      expect(selectButton.props.variant).toBe(SelectButtonVariant.Secondary);
```
georgewrmarshall: The tests in `when composing SelectButton` use `root.findByProps` and assert against `.props.accessible`, `.props.size`, and `.props.variant` directly. Per the testing rules, this checks implementation details rather than observable behavior — it couples tests to internal prop wiring rather than the rendered contract.

For `pointerEvents none` + `accessible false`, the observable outcome is that pressing the SelectButton area still fires the outer `onPress` (not a nested press on the button itself) — that's already covered by the press test above. For `size` and `variant`, the observable effect is the SelectButton's rendered visual — style assertions with `toHaveStyle` would express the contract without coupling to prop internals.

non-blocking: consider whether these three tests add coverage that isn't already captured elsewhere, and if so, rewrite them against rendered output rather than prop values.

georgewrmarshall (commented):
Solid new component — the ADR-0004 layered type structure is correct, the shared/platform split is clean, and the SelectButton composition pattern (pointerEvents none, fixed Md + Secondary) reads well. Two things worth addressing before merge: the component is missing `forwardRef`, which is required by the component creation checklist, and the "composing SelectButton" test suite asserts internal props directly rather than observable behavior.

packages/design-system-shared/src/types/KeyValueSelect/KeyValueSelect.types.ts
```diff
@@ -0,0 +1,36 @@
...
+  KeyValueRowPropsShared,
+  'value'
+> & {
+  /**
+   * Selected select label. When `undefined` or `null`, `selectButtonProps.placeholder` is shown.
+   */
+  value?: string | null;
+  /**
+   * SelectButton-only props (placeholder and caret controls).
+   */
```
georgewrmarshall, line 43: The `variant` prop is typed as `KeyValueRowVariant`, which means consumers importing `KeyValueSelect` need to reach into a sibling component's const. Per the component-scoped const naming rule (see component-architecture.md), we need a `KeyValueSelectVariant` alias in shared:

```ts
export const KeyValueSelectVariant = KeyValueRowVariant;
export type KeyValueSelectVariant = KeyValueRowVariant;
```

Then `variant?: KeyValueSelectVariant` on the shared props, and stories/README updated to import and use `KeyValueSelectVariant` throughout.

georgewrmarshall (commented):
Looking at this component more closely raised some questions for me about the API design and whether we're introducing the right primitive here.

packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+      const baseStyle = tw.style(rootTwClassName, pressed && 'bg-pressed');
+
+      if (!style) {
+        return baseStyle;
+      }
+
+      const userStyle =
+        typeof style === 'function' ? style({ pressed }) : style;
+
+      return [baseStyle, userStyle];
```
georgewrmarshall, line 70: Looking at the composition here — `KeyValueSelect` is an outer `Pressable` wrapping `KeyValueRow` with a locked-down `SelectButton` (fixed Md + Secondary, pointerEvents none) as the value. This raises a question about whether the right fix is a new composer component or adding `onPress` to `KeyValueRow` itself.

`KeyValueRow` currently renders a `BoxRow` → `View` and spreads `...viewProps` onto it. If we made `KeyValueRow` optionally pressable (accepting `onPress` and swapping to a `Pressable` root when provided), consumers get the same pressable-row behaviour with any value node:

```tsx
// pressable with SelectButton value
<KeyValueRow
  keyLabel="Network"
  value={<SelectButton pointerEvents="none" ... />}
  onPress={openSheet}
/>

// pressable with plain text value
<KeyValueRow
  keyLabel="Network"
  value="Ethereum Mainnet"
  onPress={navigate}
/>
```

That covers both cases this component solves today — plus any future value widget — without a dedicated `KeyValueSelect` wrapper. Worth discussing whether `KeyValueSelect` should be reconsidered in favour of extending `KeyValueRow`, or whether the SelectButton-specific prop mapping (`valueStartAccessory` → `startAccessory`, `valueTextProps` → `textProps`, etc.) is valuable enough to keep the composer.

packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+        textProps={valueTextProps}
+        variant={SelectButtonVariant.Secondary}
+        isDisabled={isDisabled}
+        size={SelectButtonSize.Md}
+        pointerEvents="none"
+        accessible={false}
+      />
+    );
+
+    return (
```
georgewrmarshall, line 88: There is no way for consumers to reach the inner `KeyValueRow` — no `keyValueRowProps` pass-through and no ref forwarding to it. If someone needs to set a `testID` on the row content, override row styling, or attach an accessibility label to the row separately from the outer pressable, they have no path. At minimum a `keyValueRowProps` escape hatch would unblock that without redesigning the API.

packages/design-system-react-native/src/components/KeyValueSelect/KeyValueSelect.tsx
georgewrmarshall: Worth noting how we handle this in `ListItem` — it accepts an `isInteractive` prop that swaps the root from a `View` to a `Pressable` (with `bg-pressed` feedback) internally, so the component stays a single primitive whether or not it's tappable. The same pattern on `KeyValueRow` would let consumers do this without a separate component

```tsx
<KeyValueRow
  keyLabel="Network"
  value={<SelectButton pointerEvents="none" ... />}
  isInteractive
  onPress={openSheet
/>
```
  
That covers the SelectButton case today, plus any future value widget, without `KeyValueSelect` needing to own the press behaviour. Could be worth revisiting whether this component should exist at all, or whether extending `KeyValueRow` with `isInteractive`/`onPress` is the right fix.

georgewrmarshall (approved):
All three blocking concerns from the previous review are resolved: `forwardRef` is implemented and tested, `KeyValueSelectVariant` is a proper component-scoped alias in shared, `keyValueRowProps` provides the escape hatch, and the test suite no longer asserts internal prop wiring — it uses `toBeDisabled`, `toHaveStyle`, and `toBeOnTheScreen`. Good to merge.

PR #1417, opened 2026-07-24
ci: run build, lint, and test only for changed packages

georgewrmarshall:
## **Description**

Ports the incremental CI pattern from [`MetaMask/core#9373`](https://github.com/MetaMask/core/commit/007a6187d36fad40e46625e9c5d067c5ad701d3f) and [`MetaMask/core#9643`](https://github.com/MetaMask/core/commit/baa65b0edf23311c7e6fc4511a423539c5c26d9d) to this monorepo, as suggested by Maarten Zuidhoorn in [#metamask-core-platform](https://consensys.slack.com/archives/C094GV3E7SB/p1784911909137719).

Currently every CI run rebuilds and tests all packages regardless of what changed. This PR scopes the expensive jobs to only the packages touched by a PR, plus their transitive dependants, with a full run as a safe fallback when root-level config files or `yarn.lock` change.

**`scripts/get-changed-workspaces.ts`** — New script that:
- Uses the GitHub Compare API to resolve the exact merge base between the PR head and the target branch (needed because the checkout action creates a shallow clone where `git merge-base` cannot resolve correctly)
- Runs `git diff --name-only` to find changed files
- Maps changed files to their workspace packages
- Expands the set to transitive dependants via a BFS walk of the reverse dependency graph (so a `design-system-shared` change also re-tests `design-system-react`, `design-system-react-native`, the presets, etc.)
- Falls back to a full run when root-level files change (workflow configs, tsconfigs, scripts) or `yarn.lock` changes

**`.github/workflows/lint-build-test.yml`** — Updated to:
- Compute the merge base and changed workspaces in the `prepare` job
- Scope ESLint to changed directory paths on incremental runs (other lint checks always run in full since they are fast)
- Filter the `validate-changelog` and `test` matrix jobs to only changed packages, with an early-exit guard when nothing changed (GitHub Actions errors on an empty matrix without this guard)

Key simplification vs core: Core implements smart lockfile diffing via `@yarnpkg/core` to handle `yarn.lock` changes without triggering a full run. With only 6 packages here, that complexity is not worth it — a `yarn.lock` change simply falls back to a full run.

## **Related issues**

Fixes: [DSYS-969](https://consensyssoftware.atlassian.net/browse/DSYS-969)

## **Manual testing steps**

1. Open a PR that only touches files in `packages/design-system-react-native` and confirm only that package's tests and changelog validation run
2. Open a PR that touches `packages/design-system-shared` and confirm tests expand to all dependant packages (`design-system-react`, `design-system-react-native`, the presets)
3. Open a PR that touches a root-level file (e.g. a workflow config) and confirm a full run is triggered
4. Merge to `main` and confirm a full run is triggered (no merge base available on push)

### **Before**

All packages built, linted, and tested on every PR regardless of what changed.

### **After**

Only changed packages and their transitive dependants are tested. Full run still triggered when root config or `yarn.lock` changes.

- [ ] I've included tests if applicable
- [ ] I've documented my code using [JSDoc](https://jsdoc.app/) format if applicable

---

> [!NOTE]
> **Medium Risk**
> Incorrect change detection could skip tests or lint for affected packages; fallbacks for lockfile and root/CI changes mitigate most cases.
> 
> **Overview**
> Adds **incremental CI** so PR runs focus on workspaces touched by the diff (plus packages that depend on them), instead of always linting and testing every package.
> 
> A new **`scripts/get-changed-workspaces.ts`** compares the merge base to the PR head, maps changed files to Yarn workspaces, walks a reverse dependency graph to include transitive dependants, and emits package names plus ESLint directory paths. It **falls back to a full run** when `yarn.lock` changes or when root/CI-impacting paths change (root config, `scripts/`, `.github/workflows/`).
> 
> The **`prepare` job** resolves merge base via the GitHub Compare API (for shallow checkouts), runs that script, and exposes `package-names` and `changed-paths`. **Lint** runs full `yarn lint` on fallback; otherwise ESLint is scoped to changed directories while misc/constraints/dependency checks still run globally. **`validate-changelog`** and **`test`** matrix jobs use the filtered package list and skip when the list is empty.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 577895e3a66bffd13f502f7c3a5a72b125a21cf3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


georgewrmarshall:
Good catch. `yarn lint:eslint` expands to `eslint . --cache`, so the paths appended via `xargs` came after the `.` and scoping never applied.

Fixed in c71b08d by using `readarray` to collect the paths into an array and invoking `eslint` directly:

```bash
readarray -t CHANGED_DIRS < <(echo "$CHANGED_PATHS" | jq -r '.[]')
yarn exec eslint --cache "${CHANGED_DIRS[@]}"
```

This passes the changed directory paths as the first positional arguments to ESLint, so it lints only those directories.

.github/workflows/lint-build-test.yml
```diff
@@ -39,7 +82,27 @@ jobs:
...
+          else
+            echo "Running ESLint on changed paths:"
+            echo "$CHANGED_PATHS" | jq -r '"- " + .[]'
+            echo ""
+            # Scope ESLint to changed directories; always run the fast checks.
+            # We invoke eslint directly rather than via `yarn lint:eslint`
+            # because that script expands to `eslint . --cache` — xargs
+            # would append paths after the `.`, so scoping never takes effect.
+            readarray -t CHANGED_DIRS < <(echo "$CHANGED_PATHS" | jq -r '.[]')
+            yarn exec eslint --cache "${CHANGED_DIRS[@]}"
```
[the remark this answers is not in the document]
georgewrmarshall: Fixed in 98818bb — added a length check so ESLint is skipped cleanly when `CHANGED_DIRS` is empty (e.g. a PR that only touches root docs).

scripts/get-changed-workspaces.ts
```diff
@@ -0,0 +1,189 @@
...
+// it has its own fast-path that falls back to a full run instead.
+const IGNORED_ROOT_FILES = new Set([
+  '.gitignore',
+  '.prettierignore',
+  'AGENTS.md',
+  'CLAUDE.md',
+  'CLAUDE.local.md',
+  'README.md',
+  'yarn.lock',
+]);
```
georgewrmarshall, line 22: `IGNORED_ROOT_FILES` lists root-level files whose changes should not trigger a full CI run — docs, editor config, AI agent instructions, etc. `yarn.lock` is intentionally included here so it does _not_ fall through to the generic root-change path below; it has its own explicit check on line 117 that forces a full run instead. The distinction matters: the generic root-change path would proceed with an empty workspace set if all changed files were in this ignore list, which is wrong for `yarn.lock` changes (a lockfile bump can silently affect any package at runtime).

scripts/get-changed-workspaces.ts
```diff
@@ -0,0 +1,189 @@
...
+    'yarn',
+    ['workspaces', 'list', '--no-private', '--json'],
+    { cwd: ROOT_WORKSPACE },
+  );
+
+  return stdout
+    .trim()
+    .split('\n')
+    .map((line) => JSON.parse(line))
+    .filter(({ location }: Workspace) => location !== '.');
```
georgewrmarshall, line 44: The root workspace (`location: "."`) is filtered out here because it is not a publishable package and has no test suite of its own. Root-level file changes that _do_ matter (e.g. shared config files) are caught separately by `hasRootFileChanged`, which triggers a full run across all packages.

scripts/get-changed-workspaces.ts
```diff
@@ -0,0 +1,189 @@
...
+
+      for (const depName of allDeps) {
+        if (dependants[depName]) {
+          dependants[depName].add(workspace);
+        }
+      }
+    }),
+  );
+
+  return { dependants };
```
georgewrmarshall, line 88: `buildDependantGraph` builds a _reverse_ dependency map: for each package, it records which other packages list it as a dependency. The shape is `{ "design-system-shared": Set([design-system-react, design-system-react-native, ...]) }` — meaning "if shared changes, also run tests for react and react-native." This is the opposite of the forward dep graph you see in `package.json`. The core repo ([`get-changed-workspaces.mts`](https://github.com/MetaMask/core/blob/main/scripts/get-changed-workspaces.mts)) builds the same structure; we ported the logic directly.

scripts/get-changed-workspaces.ts
```diff
@@ -0,0 +1,189 @@
...
+  const [changedFiles, workspaces] = await Promise.all([
+    getChangedFiles(mergeBase, headRef),
+    getAllWorkspaces(),
+  ]);
+
+  // yarn.lock changes: fall back to a full run. We don't parse the lockfile
+  // to find affected workspaces (unlike core), since we only have a handful
+  // of packages and the safe default is fine.
+  if (changedFiles.includes('yarn.lock')) {
+    return { workspaces, hasRootChange: true };
```
georgewrmarshall, line 118: This is the main simplification compared to the core repo implementation. Core parses `yarn.lock` with `@yarnpkg/core` to determine exactly which resolved packages changed (so a lockfile bump that only touched an unrelated dep does not force every package to rebuild). We skip that analysis and fall back to a full run whenever `yarn.lock` changes. With only ~6 packages in this monorepo, the cost of the extra CI runs is low, and it avoids pulling in a heavy lockfile-parsing dependency.

scripts/get-changed-workspaces.ts
```diff
@@ -0,0 +1,189 @@
...
+  while (queue.length > 0) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const workspace = queue.pop()!;
+    for (const dependant of dependants[workspace.name] ?? []) {
+      if (!result.has(dependant)) {
+        result.add(dependant);
+        queue.push(dependant);
+      }
+    }
+  }
```
georgewrmarshall, line 149: This is a breadth-first search over the reverse dependency graph. Starting from directly-changed packages, it walks outward to find every package that transitively depends on them. For example: if `design-system-shared` is changed, `design-system-react` and `design-system-react-native` get added because they depend on shared. If either of those also had dependants, those would be added in the next iteration. This ensures we never skip tests for a package that could be broken by an upstream change, even when the change is two hops away.

.github/workflows/lint-build-test.yml
```diff
@@ -12,17 +12,60 @@ jobs:
...
+        run: |
+          set -euo pipefail
+
+          PREFIXED_REF_REGEX='refs/heads/(.+)'
+          if [[ "$BASE_REF" =~ $PREFIXED_REF_REGEX ]]; then
+            BASE_REF="${BASH_REMATCH[1]}"
+          fi
+
+          MERGE_BASE=$(gh api "repos/$GITHUB_REPOSITORY/compare/$BASE_REF...$HEAD_SHA" --jq '.merge_base_commit.sha')
+          git fetch --unshallow --filter=blob:none --no-tags origin HEAD
```
georgewrmarshall, line 41: We use the GitHub Compare REST API rather than running `git merge-base` locally because the checkout action creates a shallow clone by default — shallow clones do not have enough history for `git merge-base` to resolve the correct commit. The API returns the merge base SHA from GitHub's server-side index without needing full history. We then `--unshallow` the local clone with `--filter=blob:none` (skip file contents; metadata only) so `git diff --name-only` on the next step can compare against that SHA without downloading the entire repo history. Core uses the same approach: https://github.com/MetaMask/core/blob/main/.github/workflows/lint-build-test.yml

.github/workflows/lint-build-test.yml
```diff
@@ -12,17 +12,60 @@ jobs:
...
+            else
+              CHANGED_PATHS=$(echo "$OUTPUT" | jq -c '.locations')
+            fi
+          else
+            # No merge base (e.g., push to main) — run everything.
+            PACKAGES=$(yarn workspaces list --no-private --json | jq --slurp --raw-output 'map(.name) | @json')
+            CHANGED_PATHS="full"
+          fi
+          echo "package-names=$PACKAGES" >> "$GITHUB_OUTPUT"
+          echo "changed-paths=$CHANGED_PATHS" >> "$GITHUB_OUTPUT"
```
georgewrmarshall, line 65: The script produces two distinct outputs that serve different downstream jobs. `package-names` is a JSON array of package names (e.g. `["@metamask/design-system-react", "@metamask/design-system-shared"]`) used by the `test` and `validate-changelog` matrix jobs to know which packages to actually run. `changed-paths` is either the string `"full"` (meaning run everything) or a JSON array of relative directory paths (e.g. `["packages/design-system-react", "packages/design-system-shared"]`) used by the `lint` job to scope ESLint to only the directories that changed. On a push to `main` where no PR exists there is no merge base, so both default to "all packages / full run."

.github/workflows/lint-build-test.yml
```diff
@@ -39,7 +82,31 @@ jobs:
...
+            echo "Running ESLint on changed paths:"
+            echo "$CHANGED_PATHS" | jq -r '"- " + .[]'
+            echo ""
+            # Scope ESLint to changed directories; always run the fast checks.
+            # We invoke eslint directly rather than via `yarn lint:eslint`
+            # because that script expands to `eslint . --cache` — xargs
+            # would append paths after the `.`, so scoping never takes effect.
+            readarray -t CHANGED_DIRS < <(echo "$CHANGED_PATHS" | jq -r '.[]')
+            if [[ ${#CHANGED_DIRS[@]} -gt 0 ]]; then
+              yarn exec eslint --cache "${CHANGED_DIRS[@]}"
```
georgewrmarshall, line 100: `yarn lint:eslint` expands to `eslint . --cache`, which always lints the entire repo — you cannot scope it by passing extra paths because they append _after_ the `.`. Instead we call `yarn exec eslint` (which runs the ESLint binary directly with no preset arguments) and pass the changed directories as the only positional arguments. `readarray` converts the JSON array to a bash array so the paths are passed as separate arguments rather than one space-joined string. The empty-array guard handles the case where `CHANGED_PATHS` is `[]` (e.g. a PR that only touches `README.md`) — without it, ESLint would receive no file arguments and exit with code 2.

.github/workflows/lint-build-test.yml
```diff
@@ -51,11 +118,12 @@ jobs:
   validate-changelog:
     name: Validate changelog
     runs-on: ubuntu-latest
+    if: needs.prepare.outputs.package-names != '[]'
```
georgewrmarshall, line 121: GitHub Actions treats an empty matrix as a workflow error by default — if `package-names` is `[]` and a matrix job tries to iterate over it, the entire job fails rather than being skipped. This `if` condition prevents that by short-circuiting the job entirely when no packages changed. The same guard appears on the `test` job (line 165) for the same reason. The `lint` and `build` jobs do not need it because they do not use a matrix.

georgewrmarshall:
@cursor push d9ff89ed49

georgewrmarshall:
@cursoragent rebase from main and resolve conflicts in `lint-build-test.yml`

georgewrmarshall:
@cursor push 62f88e5da8

georgewrmarshall:
@cursor push 0e575b1bea

PR #1432, opened 2026-08-05
ci: add check-release action to validate release PR titles

georgewrmarshall:
## **Description**

Ports MetaMask/core's [`check-release`](https://github.com/MetaMask/core/blob/main/.github/actions/check-release/action.yml) composite action to prevent release PRs from merging with invalid titles (e.g. `chore: Release 58.0.0`), which caused the 58.0.0 publish workflow to be skipped.

On release PRs (detected when package versions are bumped), CI now validates that the **PR title** matches the same `is-release` commit prefix patterns used on merge. It also checks for release conflicts when packages being released were modified on `main` ahead of the PR branch.

Suggested by @metamask-npm-publishers after the 58.0.0 release incident.

## **Related issues**

Follow-up to release 58.0.0 publish workflow skip.

## **Manual testing steps**

1. Verify CI passes on this PR (non-release PR — check-release step should be skipped)
2. After merge, open a release PR with an invalid title (e.g. `chore: Release X.Y.Z`) and confirm **Check release** fails
3. Confirm a release PR titled `Release X.Y.Z` passes

- [ ] Confirm `check-release` job wiring matches MetaMask/core pattern
- [ ] Confirm push-to-main workflows still pass when check-release step is skipped

### DIFF
--- a/.github/actions/check-release/action.yml
+++ b/.github/actions/check-release/action.yml
@@ -1,0 +1,265 @@
+name: Check release
+description: Check for conflicts in packages being released in this PR.
+
+inputs:
+  commit-starts-with:
+    description: "Validate that the release commit starts with a string in this comma-separated list. Use '[version]' to refer to the current release version."
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout and setup environment
+      uses: MetaMask/action-checkout-and-setup@v3
+      with:
+        is-high-risk-environment: false
+        skip-install: true
+        persist-credentials: false
+
+    - name: Get merge base
+      id: merge-base
+      shell: bash
+      env:
+        BASE_REF: ${{ github.event.pull_request.base.ref || github.event.merge_group.base_ref }}
+      run: |
+        set -euo pipefail
+
+        # Strip invalid prefix from `github.event.merge_group.base_ref`
+        # It comes prefixed with `refs/heads/`, but the branch is not checked out in this context
+        # We need to express it as a remote branch
+        PREFIXED_REF_REGEX='refs/heads/(.+)'
+        if [[ "$BASE_REF" =~ $PREFIXED_REF_REGEX ]]; then
+          BASE_REF="${BASH_REMATCH[1]}"
+        fi
+
+        MERGE_BASE=$(git merge-base HEAD "refs/remotes/origin/$BASE_REF")
+        echo "MERGE_BASE=$MERGE_BASE" >> "$GITHUB_OUTPUT"
+
+    - name: Check that root package.json is bumped for release PRs
+      shell: bash
+      id: check-root-package
+      env:
+        MERGE_BASE: ${{ steps.merge-base.outputs.MERGE_BASE }}
+      run: |
+        set -euo pipefail
+
+        mapfile -t PACKAGES < <(yarn workspaces list --json --no-private | jq --raw-output '.location + "/package.json"')
+        ANY_PACKAGE_BUMPED=false
+
+        for package in "${PACKAGES[@]}"; do
+          if ! PACKAGE_JSON_AT_BASE=$(git show "$MERGE_BASE:$package" 2>/dev/null); then
+            # Package didn't exist at the merge base; it's newly added, not a
+            # version bump.
+            continue
+          fi
+          CURRENT_VERSION=$(echo "$PACKAGE_JSON_AT_BASE" | jq -r .version)
+          NEW_VERSION=$(jq -r .version "$package")
+
+          if [[ "$NEW_VERSION" != "0.0.0" && "$NEW_VERSION" != "$CURRENT_VERSION" ]]; then
+            ANY_PACKAGE_BUMPED=true
+            package_name=$(jq -r ".name" "$package")
+            echo "📦 Package \`$package_name\` has a version bump (\`$CURRENT_VERSION\` -> \`$NEW_VERSION\`)"
+          fi
+        done
+
+        if [ "$ANY_PACKAGE_BUMPED" = "true" ]; then
+          CURRENT_ROOT_VERSION=$(git show "$MERGE_BASE:package.json" | jq -r .version)
+          NEW_ROOT_VERSION=$(jq -r .version package.json)
+
+          if [ "$NEW_ROOT_VERSION" = "$CURRENT_ROOT_VERSION" ]; then
+            echo "::error::This pull request is a release (one or more packages have a version bump), but the root package.json version was not bumped. Please bump the root package.json version."
+            exit 1
+          else
+            echo "✅ Root package.json is bumped (\`$CURRENT_ROOT_VERSION\` -> \`$NEW_ROOT_VERSION\`)"
+            echo "root-package-bumped=true" >> "$GITHUB_OUTPUT"
+          fi
+        else
+          echo "✅ No package version bumps detected; skipping root package.json check."
+        fi
+
+    - name: Check if the commit or pull request is a release
+      id: is-release
+      uses: MetaMask/action-is-release@v2
+      with:
+        commit-starts-with: ${{ inputs.commit-starts-with }}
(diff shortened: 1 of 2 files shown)


.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,265 @@
+name: Check release
```
georgewrmarshall, line 1: This composite action is ported from MetaMask/core: https://github.com/MetaMask/core/blob/main/.github/actions/check-release/action.yml.

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,265 @@
...
+        else
+          echo "✅ No package version bumps detected; skipping root package.json check."
+        fi
+
+    - name: Check if the commit or pull request is a release
+      id: is-release
+      uses: MetaMask/action-is-release@v2
+      with:
+        commit-starts-with: ${{ inputs.commit-starts-with }}
+        commit-message: ${{ github.event.pull_request.title }}
```
georgewrmarshall, line 85: This step validates the PR title, not the squash commit message. That matches how GitHub sets the merge commit from the PR title and is what would have blocked chore: Release 58.0.0 before merge. Core uses the same pattern: https://github.com/MetaMask/core/blob/main/.github/actions/check-release/action.yml#L80-L87

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,265 @@
...
+        before: ${{ steps.merge-base.outputs.MERGE_BASE }}
+        skip-checkout: true
+
+    - name: Fail if root package is bumped but not detected as release
```
georgewrmarshall, line 89: This is the guard that fails CI when package versions are bumped but the PR title does not match the is-release prefix list. The error message refers to commit message because squash merge uses the PR title as the commit subject.

.github/workflows/main.yml
```diff
@@ -48,6 +48,25 @@ jobs:
...
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Check release
+        if: github.event_name == 'pull_request'
```
georgewrmarshall, line 65: The check-release step runs only on pull_request events so push-to-main release merges still work. Core uses the same conditional on the step, not the job: https://github.com/MetaMask/core/blob/main/.github/workflows/main.yml#L102-L118

.github/workflows/main.yml
```diff
@@ -48,6 +48,25 @@ jobs:
...
+        uses: ./.github/actions/check-release
+        with:
+          commit-starts-with: 'Release [version],Release v[version],Release/[version],Release/v[version],Release `[version]`'
```
georgewrmarshall, line 68: This prefix list must stay in sync with the is-release job below (line 81). Core centralizes both via the RELEASE_COMMIT_PREFIX repo variable; we hardcode for now but could adopt that variable later to avoid drift.

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,265 @@
...
+          echo "TARGET=$(git merge-base HEAD refs/remotes/origin/main)" >> "$GITHUB_OUTPUT"
+        elif [ "$EVENT_NAME" = "merge_group" ]; then
+          echo "TARGET=$(git rev-parse HEAD^)" >> "$GITHUB_OUTPUT"
+        else
+          echo "::error::This action only supports \`pull_request\` and \`merge_group\` events."
+          exit 1
+        fi
+
+    - name: Check commits for changes in released packages
+      id: check-release
```
georgewrmarshall, line 136: Release conflict detection catches changes on main that landed after the release branch was cut. Same logic as core lines 136-198: https://github.com/MetaMask/core/blob/main/.github/actions/check-release/action.yml#L136-L198

.github/workflows/main.yml
```diff
@@ -78,6 +97,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - analyse-code
+      - check-release
```
georgewrmarshall, line 100: check-release is added to all-jobs-complete so release PR validation must pass before merge. Core wires it the same way: https://github.com/MetaMask/core/blob/main/.github/workflows/main.yml#L167-L172

.github/workflows/main.yml
```diff
@@ -78,6 +97,7 @@ jobs:
...
       - lint-build-test
```
georgewrmarshall, line 81: The push-time is-release job below still uses a pinned action-is-release SHA while this action uses v2. Both should agree on prefix matching; upgrading the push job to v2 is a possible follow-up (core uses v2 in both places).

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,265 @@
+name: Check release
+description: Check for conflicts in packages being released in this PR.
+
+inputs:
+  commit-starts-with:
+    description: "Validate that the release commit starts with a string in this comma-separated list. Use '[version]' to refer to the current release version."
+    required: true
+
+runs:
```
georgewrmarshall, line 9: This lives under actions/check-release/action.yml because GitHub requires local composite actions to be a directory containing a metadata file named action.yml (not action-check-release.yml at the parent level). The folder name is our choice; the filename and directory layout are GitHub Actions spec: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions. Core uses the same pattern: https://github.com/MetaMask/core/tree/main/.github/actions/check-release

brianacnguyen (dismissed):

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,265 @@
...
+        MERGE_BASE=$(git merge-base HEAD refs/remotes/origin/main)
+        for package in "${PACKAGES[@]}"; do
+          MAIN_VERSION=$(git show "$MERGE_BASE:$package" | jq -r .version)
+          HEAD_VERSION=$(jq -r .version "$package")
+
+          if [ "$HEAD_VERSION" != "$MAIN_VERSION" ]; then
+            package_name=$(jq -r ".name" "$package")
+            echo "📦 Package \`$package_name\` is being released (version \`$MAIN_VERSION\` -> \`$HEAD_VERSION\`)"
+            RELEASED_PACKAGES+=("$package")
+          fi
```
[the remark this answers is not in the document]
georgewrmarshall: Fixed in d7a1aeff: skip packages missing at merge base in the conflict loop, matching the guard in the root package check above.

We don't expect new packages in release PRs here (only version bumps, changelogs, and migration docs), but keeping the guard avoids a hard failure if Core's shared logic ever runs against an edge case. Same pattern as the existing check-root-package step.

.github/actions/check-release/action.yml
```diff
@@ -0,0 +1,265 @@
...
+        skip-checkout: true
+
+    - name: Fail if root package is bumped but not detected as release
+      if: github.event_name == 'pull_request' && steps.is-release.outputs.IS_RELEASE != 'true' && steps.check-root-package.outputs.root-package-bumped == 'true'
+      shell: bash
+      env:
+        COMMIT_STARTS_WITH: ${{ inputs.commit-starts-with }}
+      run: |
+        echo "::error::This pull request is a release, but the release check did not detect it as such. Please ensure that the commit message starts with one of the following prefixes: $COMMIT_STARTS_WITH."
+        exit 1
```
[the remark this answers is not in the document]
georgewrmarshall: Fixed in d7a1aeff: the error now tells authors to set the PR title and squash merge commit message, and explicitly warns against conventional-commit prefixes like chore:.

That matches what action-is-release validates (github.event.pull_request.title) and is the exact failure mode from release 58.0.0. Core still uses the old commit-message wording; we are correcting it here on purpose.

brianacnguyen (approved):

PR #1446, opened 2026-08-12
feat: add showCloseButton option to imperative toast API

amandaye0h:
## **Description**

Mobile consumers need toasts without a close button (e.g. watchlist confirmations), but the imperative `toast(...)` API always wired `onClose` through `Toaster`, so the X could not be hidden.

This adds `showCloseButton?: boolean` to `ToastOptions` (default `true`) for both React and React Native. When `false`, `Toaster` omits `onClose` so the close button does not render; swipe (RN), auto-dismiss, and `toast.dismiss()` still work. Docs, RN Storybook (`ShowCloseButton`), and unit tests are updated.

### Visuals

For context, the legacy component-library toast has this behavior. This is the intended outcome. Currently, it cannot be recreated.

https://github.com/user-attachments/assets/88814f2a-0458-4ef9-824b-278a7501bde6

### API usage
The mobile repo always uses the imperative API `(toast(...) + <Toaster />)`. Direct `<Toast />` is not a substitute for that flow, so close-button visibility has to be controllable via `ToastOptions`.

Screenshot 2026-08-12 at 9 59 21 PM

## **Related issues**

Fixes:

## **Manual testing steps**

1. Run `yarn storybook:ios` (or Android) and open **Components/Toast → ShowCloseButton**.
2. Tap **With close button** — confirm the X is visible and dismisses the toast.
3. Tap **Without close button** — confirm no X; dismiss via swipe and/or auto-timeout.
4. (Web) Confirm `toast({ showCloseButton: false, ... })` hides the close button while auto-dismiss / `toast.dismiss()` still work.

### **Before**

Toasts always show closed button when I tried to migrate it. 

https://github.com/user-attachments/assets/4c3ec8f4-9f9f-483b-8c54-9b44fec2c143

### **After**

https://github.com/user-attachments/assets/9e368065-b0d7-4469-8ab5-943bb6f8019d

Consumers should call this behavior this way:

Screenshot 2026-08-12 at 10 04 53 PM

- [ ] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs)
- [ ] I've completed the PR template to the best of my ability

### DIFF
--- a/packages/design-system-react-native/src/components/Toast/Toast.stories.tsx
+++ b/packages/design-system-react-native/src/components/Toast/Toast.stories.tsx
@@ -131,3 +131,36 @@ export const ActionButtonOnPress: Story = {
     title: 'Privacy policy update',
   },
 };
+
+export const ShowCloseButton: Story = {
+  render: () => (
+    <>
+      <Box twClassName="gap-2">
+        <Button
+          onPress={() => {
+            toast({
+              description: 'showCloseButton defaults to true.',
+              severity: ToastSeverity.Success,
+              title: 'Close button visible',
+            });
+          }}
+        >
+          With close button
+        </Button>
+        <Button
+          onPress={() => {
+            toast({
+              description: 'Dismiss via swipe or auto-timeout.',
+              severity: ToastSeverity.Success,
+              showCloseButton: false,
+              title: 'Close button hidden',
+            });
+          }}
+        >
+          Without close button
+        </Button>
+      </Box>
+      <Toaster />
+    </>
+  ),
+};
--- a/packages/design-system-react-native/src/components/Toast/Toast.types.ts
+++ b/packages/design-system-react-native/src/components/Toast/Toast.types.ts
@@ -31,6 +31,12 @@ export type ToastSharedProps = Omit<BannerBaseProps, 'closeButtonProps'> & {
 export type ToastOptions = ToastSharedProps & {
   hasNoTimeout?: boolean;
   topOffset?: number;
+  /**
+   * When `false`, hides the close button on the imperative toast.
+   * Swipe-to-dismiss, auto-dismiss, and `toast.dismiss()` still work.
+   * Defaults to `true`.
+   */
+  showCloseButton?: boolean;
 };
 
 /**
--- a/packages/design-system-react/src/components/Toast/README.mdx
+++ b/packages/design-system-react/src/components/Toast/README.mdx
@@ -62,7 +62,7 @@ const Demo = () => {
 
 Call `toast.dismiss()` to dismiss the currently visible toast.
 
-Use `closeButtonProps` to access the close button element when you need to set a `data-testid` or override its accessibility label. With the `toast(...)` / `<Toaster />` flow the close button is shown because `<Toaster />` wires `onClose` internally; for direct `<Toast />` rendering, provide `onClose` to render it.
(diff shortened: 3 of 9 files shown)


packages/design-system-react-native/src/components/Toast/Toast.types.ts
```diff
@@ -31,6 +31,12 @@ export type ToastSharedProps = Omit<BannerBaseProps, 'closeButtonProps'> & {
 export type ToastOptions = ToastSharedProps & {
   hasNoTimeout?: boolean;
   topOffset?: number;
+  /**
+   * When `false`, hides the close button on the imperative toast.
+   * Swipe-to-dismiss, auto-dismiss, and `toast.dismiss()` still work.
+   * Defaults to `true`.
+   */
+  showCloseButton?: boolean;
```
georgewrmarshall, line 39: non-blocking: Same design question as in the summary. For short confirmation toasts like the watchlist examples, is hiding the close button actually desired, or is the goal mainly parity with the legacy toast? Defaulting to always showing the close control (and documenting this as an escape hatch) would better match how we treat toasts / bottom sheets elsewhere.

packages/design-system-react-native/src/components/Toast/Toast.stories.tsx
```diff
@@ -131,3 +131,36 @@ export const ActionButtonOnPress: Story = {
     title: 'Privacy policy update',
   },
 };
+
+export const ShowCloseButton: Story = {
```
georgewrmarshall, line 135: non-blocking: Thanks for adding the RN story. Worth mirroring a `ShowCloseButton` story on the React package too for Storybook / Chromatic parity.

packages/design-system-react/src/components/Toast/README.mdx
```diff
@@ -179,6 +179,7 @@ toast({
 - `children`, `childrenWrapperProps` - Optional extra content rendered below the description.
 - `actionButtonLabel`, `actionButtonOnClick`, `actionButtonProps` - Optional action button content and handler.
 - `onClose` - Optional callback invoked when the close button is clicked. Also renders the close button on direct `<Toast />`.
+- `showCloseButton` - Imperative `toast(...)` only. When `false`, hides the close button. Defaults to `true`.
```
georgewrmarshall, line 182: non-blocking: The options list and intro paragraph cover `showCloseButton`, but the `### onClose` section still says the imperative flow always shows the close button because `<Toaster />` passes `onClose` internally. Worth updating that section so it does not contradict this option.

packages/design-system-react/src/components/Toast/Toast.types.ts
```diff
@@ -46,6 +46,12 @@ export type ToastOptions = ToastProps & {
    * Defaults to false.
    */
   hasNoTimeout?: boolean;
+  /**
+   * When `false`, hides the close button on the imperative toast.
+   * Auto-dismiss and `toast.dismiss()` still work.
+   * Defaults to `true`.
+   */
+  showCloseButton?: boolean;
```
georgewrmarshall, line 54: non-blocking: On web there is no swipe-to-dismiss, so `showCloseButton: false` combined with `hasNoTimeout: true` leaves only `toast.dismiss()` as an escape hatch. Consider a short docs note calling that out for extension / web consumers.

georgewrmarshall (approved):
Implementation looks correct: Toaster owns dismiss for the imperative API, so a dedicated `showCloseButton` flag (default `true`) is the right way to opt out of wiring `onClose` without a breaking change.

non-blocking: On the product/design side, the watchlist examples in the PR description ("Added to Watchlist" / "Removed from Watchlist") still look like cases where a close button would be fine. My understanding has been that toasts (and bottom sheets) always expose a close control as the default pattern. Is there a stronger design reason these need the X hidden beyond matching the legacy toast, or should we prefer keeping the close button and treat this flag as a rare escape hatch?

Screenshot 2026-08-12 at 11 47 28 AM
Screenshot 2026-08-12 at 11 47 33 AM

PR #1463, opened 2026-08-27
feat: add numeric aliases for Tailwind's radius scale

gettelmm:
## **Description**

Radius is the one visual domain with no design token behind it. Colours, typography, shadows and animations all flow through `@metamask/design-tokens`; radius does not, so components fall back to Tailwind's names directly.

**We are not introducing a scale.** Tailwind's values are already the right ones — checked, and every value the design system uses is one Tailwind already ships. What this PR adds is a naming convention on top of them, plus the CSS-variable and JS surfaces that non-Tailwind callers need.

### Numeric aliases

| Preferred | Tailwind equivalent | Value |
| --- | --- | --- |
| `rounded-none` | `rounded-none` | 0 |
| `rounded-2` | `rounded-sm` | 2px |
| `rounded-4` | `rounded` | 4px |
| `rounded-6` | `rounded-md` | 6px |
| `rounded-8` | `rounded-lg` | 8px |
| `rounded-12` | `rounded-xl` | 12px |
| `rounded-16` | `rounded-2xl` | 16px |
| `rounded-24` | `rounded-3xl` | 24px |
| `rounded-full` | `rounded-full` | 9999px |

`none` and `full` keep Tailwind's spelling — those names say something the numbers cannot, and are already the established way to write them.

### Why alias rather than just use Tailwind's names

A numeric name states its value, so a class and a comment beside it cannot drift. That is not hypothetical here: `AvatarBase` carries `rounded-sm // 4px` when Tailwind renders it at 2px.

The stronger reason is version stability. **`rounded-sm` is 2px in Tailwind v3 and 4px in v4** — the same class silently changes meaning on upgrade. `rounded-8` cannot.

### Scope

- `borderRadius` exported from `@metamask/design-tokens`, plus `--radius-*` CSS variables in `styles.css` and the Tailwind v4 `@theme`. This is the part mobile's ~508 raw `borderRadius: N` literals and extension's ~145 SCSS declarations currently have nothing to point at.
- Both presets register the numeric aliases via `theme.extend`
- A Storybook page under **Design Tokens → Border Radius**
- `twMerge` taught the aliases

### Nothing breaks, nothing needs migrating

Purely additive. Tailwind's names keep working and are not scheduled for removal — aliases and originals coexist by design. A test asserts the aliases introduce no value Tailwind does not already ship.

No `CHANGELOG.md` entries here by design: changelogs are generated on `release/*` branches, and the "Validate changelog" checks pass.

### One fix worth a look

`twMerge` had to learn the aliases. Without it, `twMerge('rounded-lg', 'rounded-12')` keeps **both** classes and CSS source order picks the winner — so a consumer override is silently ignored. Merge cannot otherwise tell the two vocabularies name the same property. The class group is keyed per corner, so all fifteen corner groups need them. Covered by tests, including alias-vs-Tailwind-name conflicts in both directions.

## **Related issues**

Fixes:

## **Manual testing steps**

1. `yarn build && yarn test && yarn lint` — all green.
2. `yarn storybook` → **Design Tokens / Border Radius**. Each step renders in both themes and computed `border-radius` matches the name.
3. Confirm the aliases are aliases: on React Native, `tw.style('rounded-8')` deep-equals `tw.style('rounded-lg')`. Asserted in tests for every pair.
4. Confirm tooling resolves them: `rounded-8` passes `tailwindcss/no-custom-classname` and appears in Tailwind IntelliSense on both platforms, since both configs read the preset.

## **Screenshots/Recordings**

[Border radius scale in Storybook](https://cursor.com/agents/bc-453ce01d-ee42-42d7-9261-4ef90bfb094d/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fscreenshot_radius_scale_storybook.png)

### **Before**

No visual change — this PR adds class names and does not alter any rendered radius. The new Storybook page is the only new UI.

### **After**

As above.

---

### What this PR is really asking

One question: **is the numeric alias the convention we want?** Everything downstream follows from it.

Chromatic will show three new snapshots for the Border Radius page and no changes to existing ones. `chromatic --exit-zero-on-changes` means CI stays green either way, so the new baselines still need accepting.

### Note on an earlier revision

This PR previously proposed a MetaMask-owned scale that replaced Tailwind's, including a 10px step. That turned out to be nine renames of values Tailwind already had, to add one step that was not real — at a cost of ~815 call sites across extension and mobile. Reworked to aliases only. The follow-up in [#1464](https://github.com/MetaMask/metamask-design-system/pull/1464) moves the design system's own components across; consumers never have to.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-453ce01d-ee42-42d7-9261-4ef90bfb094d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-453ce01d-ee42-42d7-9261-4ef90bfb094d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.cursor/rules/component-creation.md
+++ b/.cursor/rules/component-creation.md
@@ -147,7 +147,7 @@ export type { MyComponentProps } from './MyComponent.types';
 
 - ✅ Replace template div/View with Box primitive
 - ✅ Use Text component (not raw span/Text)
-- ✅ Use design token const objects (BoxBackgroundColor, BoxBorderRadius, TextVariant)
+- ✅ Use design token const objects (BoxBackgroundColor, TextVariant) and radius token classes (`rounded-8`)
 - ✅ Forward refs using `forwardRef`
 - ✅ Set displayName: `MyComponent.displayName = 'MyComponent';`
 
@@ -159,7 +159,7 @@ export const MyComponent = forwardRef<RefType, MyComponentProps>(
     <Box
       ref={ref}
       backgroundColor={BoxBackgroundColor.BackgroundDefault}
-      borderRadius={BoxBorderRadius.Md}
+      className="rounded-6"
       {...props}
     >
       <Text variant={TextVariant.BodyMd}>{children}</Text>
--- a/.cursor/rules/styling.md
+++ b/.cursor/rules/styling.md
@@ -20,10 +20,9 @@ Component styling using design tokens, Tailwind CSS (web), and TWRNC (React Nati
 
 // ✅ Correct
 <Box
-  borderRadius={BoxBorderRadius.Lg}
   backgroundColor={BoxBackgroundColor.BackgroundAlternative}
   p={4}
-  className="flex-1" // No flex-1 prop exists, so className is fine
+  className="rounded-8 flex-1" // No radius or flex-1 prop exists, so className is fine
 >
   <Text variant={TextVariant.BodyMd}>Content</Text>
 </Box>
@@ -39,10 +38,9 @@ Component styling using design tokens, Tailwind CSS (web), and TWRNC (React Nati
 
 // ✅ Correct
 <Box
-  borderRadius={BoxBorderRadius.Lg}
   backgroundColor={BoxBackgroundColor.BackgroundAlternative}
   p={4}
-  twClassName="flex-1" // No flex-1 prop exists, so twClassName is fine
+  twClassName="rounded-8 flex-1" // No radius or flex-1 prop exists, so twClassName is fine
 >
   <Text variant={TextVariant.BodyMd}>Content</Text>
 </Box>
@@ -50,9 +48,9 @@ Component styling using design tokens, Tailwind CSS (web), and TWRNC (React Nati
 
 ### Design Tokens Only
 
-- **ALWAYS** use design token generated classes (bg-default, text-error-default)
+- **ALWAYS** use design token generated classes (bg-default, text-error-default, rounded-8)
 - **NEVER** use default Tailwind colors (bg-blue-500, text-gray-700)
-- **NEVER** use arbitrary values (bg-[#4459ff], p-[16px]) unless critical
+- **NEVER** use arbitrary values (bg-[#4459ff], p-[16px], rounded-[10px]) unless critical
 - **ALWAYS** use Text component for typography, never Tailwind text classes
 
 **React Web:**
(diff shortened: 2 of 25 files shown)


amandaye0h:
Nice work. I think this makes it much more intuitive for designers and devs to understand.

Can you clarify whether this is a strong need-to-have or more of a nice-to-have for the rebrand? For this to be meaningful update, we’d need to migrate all of the existing tokens in Mobile to these new tokens, otherwise we risk adding more tech debt. I think it’s worth refining our tokens, but given the amount of other work happening, it may make sense to treat this as a separate initiative.

From my understanding, it's possible to update our button styles from square to round without introducing a whole new set of tokens. Even though Tailwind's classes may not be super intuitive, I think they're good enough for now.

Will let @georgewrmarshall @andrewjcohen share thoughts

georgewrmarshall (commented):
Thanks for putting this together @gettelmm! I agree with the intent: designers, engineers, and agents should share a clear border-radius language.

My concern is that we are choosing a new engineering vocabulary before understanding the wider system.

### Current scale comparison

| Value | Tailwind v3, used today | Tailwind v4 | Proposed alias | Internal Figma variable | Public Figma variable |
|---:|---|---|---|---|---|
| 0px | `rounded-none` | `rounded-none` | `rounded-none` | None found | None |
| 2px | `rounded-sm` | `rounded-xs` | `rounded-2` | None found | None |
| 4px | `rounded` | `rounded-sm` | `rounded-4` | `_radii/sm` | None |
| 6px | `rounded-md` | `rounded-md` | `rounded-6` | None found | None |
| 8px | `rounded-lg` | `rounded-lg` | `rounded-8` | `_radii/md` | None |
| 12px | `rounded-xl` | `rounded-xl` | `rounded-12` | `_radii/lg` | None |
| 16px | `rounded-2xl` | `rounded-2xl` | `rounded-16` | `_radii/xl` | None |
| 24px | `rounded-3xl` | `rounded-3xl` | `rounded-24` | `_radii/xxl` | None |
| Full | `rounded-full` | `rounded-full` | `rounded-full` | `_radii/round`, 999px | None |

The Figma variables are private. The collection and variables have `hiddenFromPublishing` enabled. Components can use them internally, but designers cannot select them when creating designs.

This means we do not currently have a public Figma radius scale to align this work with.

### Do the aliases solve a proven problem?

Standard Tailwind names are widely documented and already familiar to coding agents. If asked for a 24px radius, I would expect an agent to choose `rounded-3xl`.

A repository-specific class such as `rounded-24` requires extra context. We would need to maintain that context across Tailwind, TWRNC, prompts, documentation, merge configuration, and future tooling.

We should test the agent assumption before creating a new vocabulary.

Screenshot 2026-08-28 at 11 53 07 AM

The proposal also keeps the original classes. That gives humans and agents two valid names for every value. Without a clear migration or enforcement plan, this may reduce consistency instead of improving it.

### Existing usage needs an audit

A broad scan found substantial usage outside the proposed scale:

- Mobile has 626 numeric radius declarations across 42 values. About 67% match this scale.
- Extension SCSS has 153 numeric declarations across 28 values. About 69% match this scale.
- Common values outside the scale include 10px and 20px.
- Arbitrary Tailwind values also exist, including 3px, 5px, 10px, 14px, and 20px.

These counts need a semantic audit. Some values will represent circles, pills, illustrations, or one-off shapes. Others may be legitimate missing system values or design drift.

The migration in #1464 demonstrates this risk. It changes `AvatarBase` from 10px to 12px, creating a visual change that needs design approval. Renaming classes alone does not resolve existing outliers.

### Suggested next step

I suggest we:

1. Audit radius usage across Figma, Mobile, Extension, and the design-system packages.
2. Classify outliers as intentional, missing tokens, or drift.
3. Publish an intentional radius collection in Figma.
4. Decide which names form the shared language. (I suggest the Tailwind defaults as much as possible)
5. Map that language to Tailwind and TWRNC from one source of truth.
6. Add guidance or checks that discourage unsupported arbitrary values.

We may find that publishing Figma variables which map to the existing Tailwind scale solves most of the problem with less ongoing work.

If the audit shows that numeric names give enough value, I would support them. In that case, I think we should introduce them across Figma and code together, ideally generated from one token source.

I am supportive of the direction. I just think the shared language and audit should come before the implementation choice.

georgewrmarshall:
Closing for now

PR #1471, opened 2026-08-31
fix(react-native): stop BadgeWrapper infinite layout loop

juanmigdr:
## **Description**

`BadgeWrapper` positioned the badge using its own measured width and height (via `onLayout`), to center it on the anchor's corner. The problem is that applying that position changes the badge's layout by a tiny sub-pixel amount, which can change how the next measurement rounds. On certain sizes this started a loop: measure, reposition, re-measure at a slightly different size, reposition again, forever.

We found this while chasing a JS thread performance issue in MetaMask Mobile. After sign-in, CPU usage stayed pegged at 90-100% and never settled. Tracing it down led to a notification badge inside `BadgeWrapper` calling `setState` continuously, dozens of times per second, and never stopping. Logging the raw values from `onLayout` showed the width and height literally swapping back and forth between two numbers on every call (for example `12.19 / 11.81` then `11.81 / 12.19`), which confirmed it wasn't real content resizing, it was the position feedback loop described above.

The first version of this fix rounded the measured size and skipped the state update when it hadn't changed, to break the loop. While validating it against a follow-up issue (the badge would render once at the wrong spot, then visibly jump into place a frame later, since the real measurement only arrives after the first paint), it became clear the underlying measurement approach was the real source of both problems. So instead of patching around it, this PR removes the measurement entirely.

The badge is now positioned purely through layout: percentage insets (`top`/`right`/`bottom`/`left`) plus a percentage-based `transform: translate`, both resolved by the native layout engine in a single pass. There's nothing to measure, so there's nothing to jump and nothing that can loop. This also let us replace the old `0.1464` corner-offset literal with a constant derived from the actual circle-in-square geometry, instead of a magic rounded number.

### How long has this been around?

The underlying bug has been in `BadgeWrapper` since it was first added, back in March 2025. It just never got triggered before because it only happens once a badge actually appears while mounted, at a size where the rounding happens to be unstable. Mobile started using this `BadgeWrapper` for things like the wallet notification badge and the Earn asset badges in March 2026, and only noticed the CPU impact recently once that code path got hit under the right conditions.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/ASSETS-3928

## **Manual testing steps**

1. Render a `BadgeWrapper` with a badge that becomes visible after some async condition (e.g. a notification count fetched after sign-in).
2. Before this fix: watch the JS thread / CPU usage after the badge appears, it stays elevated and never settles, and the badge visibly jumps a few pixels into place after first appearing.
3. After this fix: CPU usage settles back down shortly after the badge appears, and the badge renders directly in its final position with no jump.

### **Before & After**

image

https://github.com/user-attachments/assets/b13ffcf3-81e2-4682-b633-b4a7fefcdf05

### DIFF
--- a/packages/design-system-react-native/src/components/BadgeWrapper/BadgeWrapper.test.tsx
+++ b/packages/design-system-react-native/src/components/BadgeWrapper/BadgeWrapper.test.tsx
@@ -649,6 +649,195 @@ describe('BadgeWrapper', () => {
     expect(badgeContainer.props.style[1]).toStrictEqual(expectedBadgeStyle[1]);
   });
 
+  it('ignores a subsequent onLayout call whose raw size rounds to the same integer as before', () => {
+    // Regression test for an infinite measure -> reposition -> re-measure
+    // loop: `finalPositions` derives the badge container's own position from
+    // `badgeWidth`/`badgeHeight`, and that position is applied to the same
+    // `View` that `onLayout` measures. Repositioning by a sub-pixel amount
+    // can change how the native layout engine rounds the next measured
+    // size, so unrounded raw values could previously flip-flop forever
+    // between two adjacent sub-pixel values (e.g. 12.19/11.81 swapping with
+    // 11.81/12.19), triggering `setState` (and a real reposition) on every
+    // single layout pass and pegging the JS thread. These two raw sizes are
+    // real values captured from that infinite loop; both round to (12, 12),
+    // so the second call must be a no-op and the position must still
+    // reflect the rounded size from the first layout.
+    const TestComponent = () => (
+      <BadgeWrapper
+        testID="wrapper"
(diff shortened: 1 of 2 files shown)


georgewrmarshall (commented):
Thanks Juan for finding and solving the underlying layout loop!

I ended up resolving it through a different approach in [#1474](https://github.com/MetaMask/metamask-design-system/pull/1474). That change removes layout measurement and state updates, uses static percentage positioning, and further simplifies `BadgeWrapper`.

Given the failing CI checks and Bugbot findings on this PR, #1474 should be treated as the replacement implementation.

juanmigdr:
Superseded by https://github.com/MetaMask/metamask-design-system/pull/1474

PR #1494, opened 2026-09-08
feat: use fully rounded pill shape for ButtonBase and ButtonIcon

georgewrmarshall:
## **Description**

`ButtonBase` and `ButtonIcon` now use a fully rounded border radius (`rounded-full`) on both web and mobile, so every size and variant renders as a pill. Small and medium `ButtonBase` also pick up the 16px horizontal padding that large already used.

Previously `ButtonBase` picked its radius from a size map (`rounded-lg` at `Sm`, `rounded-xl` at `Md`/`Lg`) and `ButtonIcon` only rounded fully for the `Filled` and `Floating` variants while `Default` used `rounded-lg`. Both packages now apply `rounded-full` in the component class list like any other default style.

- `ButtonBase`: the size-based `TWCLASSMAP_BUTTONBASE_BORDER_RADIUS` map and `getButtonBaseBorderRadiusTwClass` helper were removed. Neither was exported from the package entry point, so this is not a public API change.
- `ButtonIcon`: the per-variant radius branching was removed.
- `SegmentedControl` (React Native): its container radius was sized to match the old segment radius (`rounded-xl` / `rounded-2xl`), which left the pill segments no longer concentric with the container. The container is now `rounded-full`, and its own `TWCLASSMAP_SEGMENTEDCONTROL_BORDER_RADIUS` map and `getSegmentedControlBorderRadiusTwClass` helper were deleted. The result is exactly concentric at every size: a `Sm` control is 42px tall (32px segment + 4px padding + 1px border per side) for a 21px outer radius, matching the 16px segment radius plus the 5px inset.

### Horizontal padding

Small and medium used a flat 12px inset on every side while large used 16px for a label and pulled in to 12px on any side holding an accessory. Padding no longer depends on size at all, so every size gets the 16px label inset and the same accessory rule, which balances better against the fully rounded ends:

| Accessories | Padding |
| ----------- | ------- |
| Label only | 16px both sides (`px-4`) |
| Start icon | 12px leading, 16px trailing (`pl-3 pr-4`) |
| End icon | 16px leading, 12px trailing (`pl-4 pr-3`) |
| Both | 12px both sides (`px-3`) |

Large is unchanged. `getButtonBaseHorizontalPaddingTwClasses` no longer takes a size argument. Trailing insets that downstream components compensate for are unaffected: `SelectButton` renders an end arrow by default, so its trailing padding stays 12px and `KeyValueSelect`'s `pr-1` still totals 16px from the row edge.

[...]

PR #1502, opened 2026-09-09
fix(react): stop ButtonIcon shrinking in flex containers

georgewrmarshall:
## **Description**

`ButtonIcon` sets its dimensions with `h-8 w-8` but never opts out of flex shrinking. The CSS default is `flex-shrink: 1`, so whenever the button is a flex item in a row that overflows, it is the element that absorbs the overflow and stops being square. On the `Filled` and `Floating` variants, which are `rounded-full`, the squashed box renders as an ellipse rather than a circle.

This is what happens in the Extension onboarding "Create password" screen (MetaMask/metamask-extension#46163): `FormTextField` is `inline-flex`, and `.mm-text-field__input` has `width: 100%` + `flex-grow: 1`, so the input's flex basis alone fills the container. The row overflows by the width of the accessory, and the show/hide eye `ButtonIcon` shrinks from 32px to ~28px wide while staying 32px tall.

The fix adds `shrink-0` to the base classes, matching what `AvatarBase` already does for the same reason:

```
'inline-flex shrink-0 items-center justify-center p-0'
```

Because it is in the base class list and goes through `twMerge`, consumers that genuinely want a shrinkable icon button can still pass `shrink` via `className`.

This PR is deliberately scoped to that one line. It is now based on `main` rather than on the pill-shape branch, so it carries no other changes and can land independently of MetaMask/metamask-design-system#1494.

**React Native was checked and needs no change.** Yoga's default is `flexShrink: 0`, and neither `ButtonIcon` nor the underlying `ButtonAnimated` sets a shrink value, so the RN component already keeps its square dimensions inside a row.

## **Related issues**

Related: MetaMask/metamask-extension#46163

## **Manual testing steps**

1. Render a `ButtonIcon` as the end accessory of an `inline-flex` container whose input has `width: 100%` and `flex-grow: 1` (the Extension `FormTextField` layout).
2. Before this change the button measures 27.97 x 32; after it measures 32 x 32.
3. `yarn workspace @metamask/design-system-react run test` — all suites pass.

No test was added, since `shrink-0` is a default style like the rest of the base classes.

## **Screenshots/Recordings**

Reproduced in Storybook using the real `ButtonIcon` (`Filled` variant) in the Extension `FormTextField` layout, magnified 2x. The "before" row is the same component with `shrink` re-enabled via `className`, which reproduces the pre-fix computed style exactly.

image

### **Before**

27.97 x 32 — oval.

image

### **After**

32 x 32 — circle.

image

<div><a href="https://cursor.com/agents/bc-eb292a5d-20ef-4c60-a10b-8e041044a3e9?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-eb292a5d-20ef-4c60-a10b-8e041044a3e9&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/design-system-react/src/components/ButtonIcon/ButtonIcon.tsx
+++ b/packages/design-system-react/src/components/ButtonIcon/ButtonIcon.tsx
@@ -32,7 +32,7 @@ export const ButtonIcon = forwardRef<HTMLButtonElement, ButtonIconProps>(
 
     const mergedClassName = twMerge(
       // Base styles
-      'inline-flex items-center justify-center p-0',
+      'inline-flex shrink-0 items-center justify-center p-0',
       // Size styles
       TWCLASSMAP_BUTTONICON_SIZE_DIMENSION[size],
       // Variant styles


packages/design-system-react/src/components/ButtonIcon/ButtonIcon.test.tsx
```diff
@@ -23,6 +23,16 @@ describe('ButtonIcon', () => {
...
 
+  it('keeps its square dimensions inside a constrained flex container', () => {
+    render(
+      <div className="flex w-4">
+        <input className="w-full grow" />
+        <ButtonIcon iconName={IconName.Eye} ariaLabel="Show password" />
+      </div>,
+    );
+    expect(screen.getByRole('button')).toHaveClass('shrink-0', 'h-8', 'w-8');
+  });
```
georgewrmarshall: @cursoragent we don't need this test if its a default style I think this is fine to have without a test

brianacnguyen (dismissed):

georgewrmarshall:
@cursoragent this should just be for the ButtonIcon fix isolate that change to add shrink-0 only and update this PR to be only that change

brianacnguyen (approved):

PR #1506, opened 2026-09-10
fix(react): use muted background for TextField

georgewrmarshall:
## **Description**

The React `TextField` container used `bg-default`, while the React Native `TextField` and both `TextFieldSearch` implementations already used `bg-muted`. That left the same component looking different across platforms, and left a plain text field looking different from a search field on web.

This changes the React `TextField` container background to `bg-muted` so all text field surfaces match. Because `FormTextField` and `TextFieldSearch` compose `TextField`, they pick up the muted surface too. There are no API or React Native implementation changes; React Native already used `bg-muted`.

The testing guidance now clarifies that unconditional default styles should not get dedicated assertions unless coverage requires them, and that obsolete tests and documentation should be removed when conditional style logic is removed.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Run `yarn storybook` and open **React Components → TextField**.
2. Confirm the field surface is the muted background in both light and dark mode, across default, filled, accessory, error, disabled, and read-only stories.
3. Compare against **TextFieldSearch** and **FormTextField** — all three surfaces should match.
4. Open the React Native Storybook **TextField** stories and confirm they are unchanged.

## **Screenshots/Recordings**

React Storybook, light and dark, showing default, filled, accessories, small, error, disabled, `FormTextField`, and `TextFieldSearch`.

### **Before**

`TextField` (and `FormTextField`) render on `bg-default`, which does not match the muted `TextFieldSearch` below it.

Before

### **After**

All text field surfaces use `bg-muted`.

After

---

> [!NOTE]
> **Low Risk**
> Single Tailwind token swap on a presentational container with no API or logic changes.
> 
> **Overview**
> Aligns the React **`TextField`** container surface with other text field components by changing the base Tailwind class from **`bg-default`** to **`bg-muted`**. **`FormTextField`** and **`TextFieldSearch`** inherit the same look because they compose **`TextField`**; React Native is unchanged.
> 
> The testing rules add a **Style Behavior Coverage** section: avoid tests that only assert unconditional default classes, drop or update tests and docs when conditional style branches are removed, and only assert new defaults when coverage requires it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 833c039b43cd645e66eddf869390cfc1cb9073f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.cursor/rules/testing.md
+++ b/.cursor/rules/testing.md
@@ -72,6 +72,12 @@ Optional slots (accessories, labels, custom rows) are usually **strings or eleme
 - **PREFER** `toBeOnTheScreen()` for presence checks over `toBeDefined()` / `toBeTruthy()`
 - **PREFER** specific assertions (`toHaveTextContent`, `toBeDisabled`, `toHaveStyle`)
 
+### Style Behavior Coverage
+
+- **DO NOT** add tests solely to assert unconditional default classes or styles. Test them only when required for coverage.
+- **ALWAYS** remove tests for conditional style behavior when that condition is removed; do not replace them with assertions for the new unconditional default unless coverage requires it.
+- **ALWAYS** remove or update documentation that describes style logic or variants that no longer exist.
+
 ### React Web Style Assertions
 
 - **ALWAYS** use `toHaveClass()` for Tailwind class assertions
--- a/packages/design-system-react/src/components/TextField/TextField.tsx
+++ b/packages/design-system-react/src/components/TextField/TextField.tsx
@@ -103,7 +103,7 @@ export const TextField = forwardRef<HTMLDivElement, TextFieldProps>(
       inputProps ?? {};
 
     const containerClassName = twMerge(
-      'inline-flex items-center rounded-lg border bg-default transition-colors',
+      'inline-flex items-center rounded-lg border bg-muted transition-colors',
       SIZE_CLASS[size],
       startAccessory ? 'pl-4' : 'pl-0',
       endAccessory ? 'pr-4' : 'pr-0',
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/metamask-design-system",
-  "version": "66.0.0",
+  "version": "67.0.0",
   "private": true,
   "description": "The MetaMask Design System monorepo",
   "repository": {
--- a/packages/design-system-react-native/CHANGELOG.md
+++ b/packages/design-system-react-native/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.46.0]
+
+### Added
+
+- Added `HeartStraight` and `HeartStraightFilled` to `IconName` ([#1503](https://github.com/MetaMask/metamask-design-system/pull/1503))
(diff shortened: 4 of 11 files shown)


packages/design-system-react/src/components/TextField/TextField.tsx
```diff
@@ -103,7 +103,7 @@ export const TextField = forwardRef<HTMLDivElement, TextFieldProps>(
       inputProps ?? {};
 
     const containerClassName = twMerge(
-      'inline-flex items-center rounded-lg border bg-default transition-colors',
+      'inline-flex items-center rounded-lg border bg-muted transition-colors',
```
georgewrmarshall, line 106: Native TextField already uses `bg-muted` on the field chrome. This aligns the web container with that surface so the control reads as a recessed field on `bg-default` pages rather than blending into the page background. The inner input stays `bg-transparent`, so the token only needs to live on this wrapper.

.cursor/rules/testing.md
```diff
@@ -72,6 +72,12 @@ Optional slots (accessories, labels, custom rows) are usually **strings or eleme
 - **PREFER** `toBeOnTheScreen()` for presence checks over `toBeDefined()` / `toBeTruthy()`
 - **PREFER** specific assertions (`toHaveTextContent`, `toBeDisabled`, `toHaveStyle`)
 
+### Style Behavior Coverage
+
+- **DO NOT** add tests solely to assert unconditional default classes or styles. Test them only when required for coverage.
+- **ALWAYS** remove tests for conditional style behavior when that condition is removed; do not replace them with assertions for the new unconditional default unless coverage requires it.
```
georgewrmarshall, line 78: This is why the PR does not add a `bg-muted` class assertion. The token is now an unconditional default on both platforms; locking it in a unit test would only churn when the default token changes. Keep coverage for conditional styles (error, focus, disabled borders) which still branch.

georgewrmarshall, line 78: Agents tend to add coverage for default styles even when nothing branches on them. The usual failure mode is rewriting the old assertions instead of deleting them: in https://github.com/MetaMask/metamask-design-system/pull/1494, once ButtonBase used `rounded-full` at every size, agents first updated the size-based radius tests to expect the same class on Sm/Md/Lg rather than removing that suite. This rule exists so that when style logic becomes unconditional, the related tests (and docs) go away unless coverage actually requires them.

packages/design-system-react/src/components/TextField/TextField.tsx
georgewrmarshall, line 106: `bg-muted` is now the shared TextField surface. React Native already used it, and that matches Mobile's legacy TextField (`theme.colors.background.muted` in `app/component-library/components/Form/TextField/TextField.styles.ts`). Extension's legacy TextField still uses `BackgroundColor.backgroundDefault` (`ui/components/component-library/text-field/text-field.tsx`), so this is a visual change for web relative to that source. `FormTextField` and `TextFieldSearch` inherit the muted surface through composition; Search still layers `rounded-full` and hover on top. No dedicated `bg-muted` assertion is added because this is an unconditional default, not a variant or state branch.

amandaye0h (approved):

